PR #1, opened 2017-10-05
ENT-594/ENT-599 New standalone python script for enterprise data passback

brittneyexline:
Originally implemented as a management command, this is the refactored version of the script that will be invoked to generate and email enterprise reports.

The script can be invoked with an optional argument of a particular enterprise customer. It will start by calling the enterprise API to get the reporting configurations (or just one if an enterprise uuid is passed in). It will then kick off the process of querying vertica, writing to a csv file, encrypting the csv file, and emailing it to the enterprise customer in a new celery task.


brittneyexline:
I still have some work to do on this but would like some initial feedback on both things that should be in this new repo and the script implementation itself. @douglashall @haikuginger

douglashall:
@brittneyexline Looks like the right direction to me. I'm thinking this repo could eventually be the thing that houses the Django project that delivers our Enterprise Data API and dashboard.

Makefile
```diff
@@ -0,0 +1,15 @@
...
+
+upgrade: ## update the requirements/*.txt files with the latest packages satisfying requirements/*.in
+	pip install -q pip-tools
+	pip-compile --upgrade -o requirements/base.txt requirements/base.in
+
+requirements: ## install development environment requirements
+	pip install -qr requirements/base.txt --exists-action w
+	pip-sync requirements/base.txt
+
+.PHONY: requirements upgrade
```
haikuginger: Does `help` need to be a phony target?

enterprise_reporting/clients/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+        Connect to the REST API, authenticating with an access token retrieved with our client credentials.
+        """
+
+        access_token, expires_at = EdxRestApiClient.get_oauth_access_token(
+            self.API_BASE_URL + '/access_token',
+            self.client_id,
+            self.client_secret
+        )
+        self.client = EdxRestApiClient(
+            self.API_BASE_URL, append_slash=self.APPEND_SLASH, oauth_access_token=access_token,
```
haikuginger: Do we want to be getting an OAuth access token as opposed to JWT? My understanding (which could be wrong) was that edX was shifting to JWTs.

enterprise_reporting/clients/enterprise.py
```diff
@@ -0,0 +1,69 @@
...
+        :param resource_id: The resource ID for the specific detail to get from the endpoint.
+        :param querystring: Optional query string parameters.
+        :param traverse_pagination: Whether to traverse pagination or return paginated response.
+        :param default: The default value to return in case of no response content.
+        :return: Data returned by the API.
+        """
+        default_val = default if default != self.DEFAULT_VALUE_SAFEGUARD else {}
+        querystring = querystring if querystring else {}
+
+        endpoint = getattr(self.client, resource)(resource_id)
```
haikuginger: Can you make the call conditional on `resource_id`?

enterprise_reporting/clients/enterprise.py
```diff
@@ -0,0 +1,69 @@
...
+        :param resource: The endpoint resource name.
+        :param detail_resource: The sub-resource to append to the path.
+        :param resource_id: The resource ID for the specific detail to get from the endpoint.
+        :param querystring: Optional query string parameters.
+        :param traverse_pagination: Whether to traverse pagination or return paginated response.
+        :param default: The default value to return in case of no response content.
+        :return: Data returned by the API.
+        """
+        default_val = default if default != self.DEFAULT_VALUE_SAFEGUARD else {}
+        querystring = querystring if querystring else {}
```
haikuginger: Can be simplified to `querystring or {}`.

enterprise_reporting/clients/enterprise.py
```diff
@@ -0,0 +1,69 @@
...
+
+        :param resource: The endpoint resource name.
+        :param detail_resource: The sub-resource to append to the path.
+        :param resource_id: The resource ID for the specific detail to get from the endpoint.
+        :param querystring: Optional query string parameters.
+        :param traverse_pagination: Whether to traverse pagination or return paginated response.
+        :param default: The default value to return in case of no response content.
+        :return: Data returned by the API.
+        """
+        default_val = default if default != self.DEFAULT_VALUE_SAFEGUARD else {}
```
haikuginger: Can we use an explicit statement of identity (`if default is not self.DEFAULT_VALUE_SAFEGUARD`) as opposed to equality?

enterprise_reporting/clients/enterprise.py
```diff
@@ -0,0 +1,69 @@
...
+        querystring = querystring if querystring else {}
+
+        endpoint = getattr(self.client, resource)(resource_id)
+        endpoint = getattr(endpoint, detail_resource) if detail_resource else endpoint
+        response = endpoint.get(**querystring)
+        if should_traverse_pagination:
+            results = traverse_pagination(response, endpoint)
+            response = {
+                'count': len(results),
+                'next': 'None',
```
haikuginger: Should `next` and `previous` be `"None"` or `None`?

enterprise_reporting/clients/vertica.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Instantiate a new client using the Django settings to determine the vertica credentials.
+
+        If there are none configured, throw an exception.
+        """
+        self.connection_info = {
+            'host': host,
+            'user': username,
+            'password': password,
+            'database': 'warehouse',
```
haikuginger, line 30: Should we make this database name configurable by the Jenkins task?

enterprise_reporting/send_enterprise_reports.py
```diff
@@ -0,0 +1,82 @@
...
+
+@task
+def send_data_task(reporting_config):
+    """
+    Task to send data report to each enterprise.
+
+    Args:
+        reporting_config
+    """
+    enterprise_customer_name = reporting_config.enterprise_customer.name
```
haikuginger: `reporting_config` will be a dictionary, so use `.get()` instead of dot-attribute access.

enterprise_reporting/reporter.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        self.vertica_client.connect()
+
+    def send_enterprise_report(self):
+        """
+        Query the data warehouse (vertica) and export data to a csv file.
+
+        This file will get encrypted and emailed to the Enterprise Customer.
+        """
+        enterprise_customer_name = self.reporting_config.enterprise_customer.name
```
haikuginger: `reporting_config` will be a nested dict, not an object.

enterprise_reporting/send_enterprise_reports.py
```diff
@@ -0,0 +1,82 @@
...
+        os.environ.get('OAUTH_KEY'),
+        os.environ.get('OAUTH_SECRET')
+    )
+
+    if args.enterprise_customer:
+        reporting_configs = enterprise_api_client.get_enterprise_reporting_config(args.enterprise_customer)
+    else:
+        reporting_configs = enterprise_api_client.get_all_enterprise_reporting_configs()
+
+    if args.enterprise_customer and not reporting_configs.exists():
```
haikuginger: We won't be able to use the queryset `.exists()` method, since we're not in Django - just check for truthiness.

enterprise_reporting/send_enterprise_reports.py
```diff
@@ -0,0 +1,82 @@
...
+                        help="EnterpriseCustomer UUID.")
+
+    args = parser.parse_args()
+    enterprise_api_client = EnterpriseAPIClient(
+        os.environ.get('OAUTH_KEY'),
+        os.environ.get('OAUTH_SECRET')
+    )
+
+    if args.enterprise_customer:
+        reporting_configs = enterprise_api_client.get_enterprise_reporting_config(args.enterprise_customer)
```
haikuginger, line 73: Note that this will need a bit of modification, since `get_enterprise_reporting_config` is likely to return a dict as opposed to the list-like QuerySet object that we were getting inside Django.

enterprise_reporting/reporter.py
```diff
@@ -0,0 +1,117 @@
...
+        # initialize base csv file and file writer
+        data_report_file_name, data_report_file_writer = self._create_data_report_csv_writer()
+
+        # query vertica and write each row to the file
+        LOGGER.debug('Querying Vertica for data for {}'.format(enterprise_customer_name))
+        data_report_file_writer.writerows(self._query_vertica())
+
+        # create a password encrypted zip file
+        LOGGER.debug('Encrypting data report for {}'.format(enterprise_customer_name))
+        data_report_zip_name = compress_and_encrypt(data_report_file_name, self.reporting_config.password)
```
haikuginger: `reporting_config` will be a dict.

enterprise_reporting/reporter.py
```diff
@@ -0,0 +1,117 @@
...
+
+        # email the file to the email address in the configuration
+        LOGGER.debug('Sending encrypted data to {}'.format(enterprise_customer_name))
+        try:
+            send_email_with_attachment(
+                self.REPORT_EMAIL_SUBJECT,
+                self.REPORT_EMAIL_BODY,
+                self.REPORT_EMAIL_FROM_EMAIL,
+                self.reporting_config.email,
```
haikuginger: `reporting_config` will be a dict.

enterprise_reporting/reporter.py
```diff
@@ -0,0 +1,117 @@
...
+            LOGGER.exception('Failed to send email for {}'.format(enterprise_customer_name))
+
+        self._cleanup([data_report_file_name, data_report_zip_name])
+
+    def _create_data_report_csv_writer(self):
+        """
+        Create a csv file and file writer with the field headers for the data report.
+        """
+        data_report_file_name = self.REPORT_FILE_NAME_FORMAT.format(
+            enterprise_id=self.reporting_config.enterprise_customer.uuid,
```
haikuginger: `reporting_config` will be a dict.

enterprise_reporting/reporter.py
```diff
@@ -0,0 +1,117 @@
...
+        data_report_file_writer.writerow(self.VERTICA_QUERY_FIELDS)
+        return data_report_file_name, data_report_file_writer
+
+    def _query_vertica(self):
+        """
+        Use a connection to Vertica to execute the report query.
+        """
+        query = self.VERTICA_QUERY.format(
+            fields=','.join(self.VERTICA_QUERY_FIELDS),
+            enterprise_id=self.reporting_config.enterprise_customer.uuid.hex
```
haikuginger: `reporting_config` will be a dict.

enterprise_reporting/utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # the attachment
+    part = MIMEApplication(open(filename, 'rb').read())
+    part.add_header('Content-Disposition', 'attachment', filename=os.path.basename(filename))
+    msg.attach(part)
+
+    # connect to SES
+    client = boto3.client(
+        'ses',
+        aws_access_key_id=os.environ.get(''),
```
haikuginger: From what I recall from the conversation in HipChat, I don't think we need to even set these? Is that right, @fredsmith?

enterprise_reporting/utils.py
```diff
@@ -0,0 +1,101 @@
...
+    # what a recipient sees if they don't use an email reader
+    msg.preamble = 'Multipart message.\n'
+
+    # the message body
+    part = MIMEText(body)
+    msg.attach(part)
+
+    # the attachment
+    part = MIMEApplication(open(filename, 'rb').read())
+    part.add_header('Content-Disposition', 'attachment', filename=os.path.basename(filename))
```
haikuginger, line 67: I'm not sure if the MIME type for the ZIP file will get set automatically based on file extension; can we set it to `application/zip` using [`set_type`](https://docs.python.org/2/library/email.message.html#email.message.Message.set_type)?

enterprise_reporting/utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # connect to SES
+    client = boto3.client(
+        'ses',
+        aws_access_key_id=os.environ.get(''),
+        aws_access_key_secret=os.environ.get('')
+    )
+
+    # and send the message
+    result = client.send_raw_email(msg.as_string(), source=msg['From'], destinations=[msg['To']])
```
haikuginger: Right around here, we should be aware of the 10MB attachment limit for SES and see if we can potentially catch an exception produced by an oversize file.

haikuginger (commented):
Definitely a work in progress, but we're getting there! Thanks so much for working on this, @brittneyexline! I've mostly called out below some items that need adjusted logic compared to the version that was in Django.

enterprise_reporting/clients/__init__.py
brittneyexline: we are, but getting a JWT vs other token format as the access token is dependent on which type of client/secret you have configured - but requesting one all goes to the same endpoint. so as long as we configure client/secret for django oauth toolkit we should get a jwt back here. but you are right that i should pass this in as a jwt instead if that's the case.

enterprise_reporting/clients/enterprise.py
brittneyexline: sure, this is something i took out of https://github.com/edx/edx-enterprise/blob/master/enterprise/api_client/enterprise.py#L47 but i'm happy to rework

enterprise_reporting/clients/vertica.py
brittneyexline: i don't think so - it's not likely to change before we have to switch away from vertica anyway.

enterprise_reporting/reporter.py
brittneyexline: yea, i forgot to switch that over, will do.

enterprise_reporting/utils.py
brittneyexline: i was not sure which is why i put placeholders here, i'll remove if it's ok to.

enterprise_reporting/reporter.py
```diff
@@ -0,0 +1,133 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+class EnterpriseReportSender(object):
+    """
+    Class that handles the process of sending a data report to an Enterprise Customer.
+    """
+
+    VERTICA_QUERY = ("SELECT {fields} FROM business_intelligence.enterprise_enrollment"
+                     " WHERE enterprise_id = '{enterprise_id}' AND consent_granted = 1")
```
bill-filler, line 26: Is there a reason we are only including records where user has consented? Our support team is interested in seeing a report that shows all enrollments and the consent status, not just the consented users.

brittneyexline: if this going to an enterprise (which is what is supported here) we shouldn't send enrollment data about users who didn't consent.

bill-filler: I thought somewhere in the default DSC wording it says that we would share the fact that the user hasn't consented, but could be wrong.

bill-filler: either way, maybe we could add a command line option, which would be False by default, whether or not to include non-consenting users

bill-filler: Just checked, in the default Data Sharing Policy it says:
"Please note
If you decline to consent, that fact may be shared with {Enterprise Customer Name}."

douglashall: We can definitely make this configurable at some point in the future. I don't think it makes sense to do that as a part of this PR.

brittneyexline: you are right @bill-filler - though we'll have to scrub the data for declined consent rows if we want to only reflect the fact that they declined consent. the way the table is structured now it's not clear to me what would need to get scrubbed. i think this is something we should revisit as @douglashall suggests.

enterprise_reporting/clients/enterprise.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        Loads a response from a call to one of the Enterprise endpoints.
+
+        :param resource: The endpoint resource name.
+        :param detail_resource: The sub-resource to append to the path.
+        :param resource_id: The resource ID for the specific detail to get from the endpoint.
+        :param querystring: Optional query string parameters.
+        :param traverse_pagination: Whether to traverse pagination or return paginated response.
+        :param default: The default value to return in case of no response content.
+        :return: Data returned by the API.
```
douglashall: Use "google" style comments:
```
Arguments:
Returns:
```

enterprise_reporting/send_enterprise_reports.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/python3
```
douglashall, line 1: Might be good to put this under a "scripts" folder to identify it as the starting point for most of this code.

douglashall (approved):
Noting that we don't have any tests or quality checks here yet. Not sure when we should add those.

PR #3, opened 2017-10-27
ENT-599 Fixing mixup between dash and underscore

brittneyexline:
*The jenkins job is expecting the long argument to be enterprise-customer
instead of enterprise_customer, so calling with the argument fails when you
build with parameters. Fixing that here because it is easier :-)
*Adding the tpa id in the data query pull.
*Fixing reading of 'SEND_EMAIL_FROM' environment setting
*Defining the AWS region
*Fixing the SES email call

### DIFF
--- a/enterprise_reporting/reporter.py
+++ b/enterprise_reporting/reporter.py
@@ -27,6 +27,7 @@ class EnterpriseReportSender(object):
     VERTICA_QUERY_FIELDS = (
         'enterprise_user_id',
         'lms_user_id',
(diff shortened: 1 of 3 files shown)


brittneyexline:
I'm expecting that there might be more changes to make here once I get non script related issues resolved, specifically around email and SES, since that part depends on jenkins/aws environment to test.

brittneyexline:
I was able to get everything to run successfully with this branch.

douglashall (approved):

PR #7, opened 2017-11-20
[ENT-765] Report subject line now includes enterprise name

georgebabey:
screen shot 2017-11-20 at 4 21 40 pm

Verified here: https://tools-edx-jenkins.edx.org/job/enterprise/job/enterprise-data-reporting/706/

Sender name change will come PR to jenkins-job-dsl-internal repo (https://github.com/edx/jenkins-job-dsl-internal/pull/145)

### DIFF
--- a/enterprise_reporting/reporter.py
+++ b/enterprise_reporting/reporter.py
@@ -44,7 +44,7 @@ class EnterpriseReportSender(object):
         'user_current_enrollment_mode',
     )
     REPORT_FILE_NAME_FORMAT = "{path}/{enterprise_id}_{date}.{extension}"
-    REPORT_EMAIL_SUBJECT = 'edX Learner Report'
+    REPORT_EMAIL_SUBJECT = '{enterprise_name} edX Learner Data'
     REPORT_EMAIL_BODY = ''
     REPORT_EMAIL_FROM_EMAIL = os.environ.get('SEND_EMAIL_FROM')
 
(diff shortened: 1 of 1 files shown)


brittneyexline (approved):
looks good, but for the email address part of this change you'll have to change the jenkins dsl for generating these jobs. this was the original pr for that work: https://github.com/edx/jenkins-job-dsl-internal/pull/133/files
you'll probably have to get access from @fredsmith  and also get him to add business@edx.org to the list of emails we can send from in AWS

georgebabey:
@brittneyexline Thanks - I have https://github.com/edx/jenkins-job-dsl-internal/pull/145, but as you mentioned, I first need the aws resource access updated for business@edx.org (https://openedx.atlassian.net/browse/DEVOPS-6802)

enterprise_reporting/reporter.py
```diff
@@ -45,7 +45,12 @@ class EnterpriseReportSender(object):
     )
     REPORT_FILE_NAME_FORMAT = "{path}/{enterprise_id}_{date}.{extension}"
     REPORT_EMAIL_SUBJECT = '{enterprise_name} edX Learner Data'
-    REPORT_EMAIL_BODY = ''
+    REPORT_EMAIL_BODY ="""
+Please find attached employee progress data for courses on edX.
```
georgebabey: Not the prettiest way to send an email body but gets the job done - could use external templated file instead but might be overkill - thoughts @douglashall ?

douglashall: Yeah, it would be good if we had a template for this. This will get the job done though.

douglashall (approved):

enterprise_reporting/reporter.py
georgebabey: Considering the email previously had no body at all, I'm going to merge this in and we can pretty it up at another time

PR #9, opened 2017-11-29
Close the report file descriptor before creating zip file.

douglashall:
This should fix a problem we were seeing where the final report file had records truncated from the end.

### DIFF
--- a/enterprise_reporting/reporter.py
+++ b/enterprise_reporting/reporter.py
@@ -77,12 +77,18 @@ The edX for Business Team
 
         LOGGER.info('Starting process to send email report to {}'.format(enterprise_customer_name))
 
-        # initialize base csv file and file writer
-        data_report_file_name, data_report_file_writer = self._create_data_report_csv_writer()
-
-        # query vertica and write each row to the file
-        LOGGER.debug('Querying Vertica for data for {}'.format(enterprise_customer_name))
-        data_report_file_writer.writerows(self._query_vertica())
+        # Query vertica and write output to csv file.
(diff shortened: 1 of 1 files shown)


douglashall:
@georgebabey @brittneyexline Reminder to review this. It would be great if I could merge this before the 11am run of the Jenkins job.

enterprise_reporting/reporter.py
```diff
@@ -110,21 +115,6 @@ def send_enterprise_report(self):
...
-        """
-        data_report_file_name = self.REPORT_FILE_NAME_FORMAT.format(
-            path=self.FILE_WRITE_DIRECTORY,
-            enterprise_id=self.reporting_config['enterprise_customer']['uuid'],
-            date=datetime.datetime.now().strftime("%Y-%m-%d"),
-            extension='csv',
-        )
-        data_report_file = open(data_report_file_name, 'w')  # pylint: disable=open-builtin
-        data_report_file_writer = csv.writer(data_report_file)
-        data_report_file_writer.writerow(self.VERTICA_QUERY_FIELDS)
```
brittneyexline, line 125: we still need to write the first row with the field names above

georgebabey (approved):
Looks good :shipit:

enterprise_reporting/reporter.py
douglashall: Good catch. I lost that row in the copy/paste.

douglashall: Should be all set now.

PR #18, opened 2018-01-30
Fix reference to Fernet key variable.

douglashall:


georgebabey (approved):
Looks good assuming that is what is in the business-secure repo (I don't have access)  :shipit:

PR #25, opened 2018-04-20
Add yml configs for api gateway

georgebabey:


douglashall (approved):
I wonder if there is a way to refactor out the common bits of the Swagger yaml into some shared file that can be used cross-project.

georgebabey:
@douglashall Maybe - would be nice to have the responses & mapping templates for pagination, uuid handling, etc in a swagger-common repo. For this I'm intentionally trying to keep the fewest moving parts in case something blows up. We can look at refactoring once this is working

PR #26, opened 2018-04-23
Rename .yml to .yaml

georgebabey:
Matches other .yaml files used by the api gateway


douglashall (approved):

georgebabey:
> Is there an official extension for YAML files?
>
> Please use ".yaml" when possible.

http://yaml.org/faq.html

PR #27, opened 2018-04-30
[ENT-872] Push catalogs to direct integration customers in CSV format (SFTP, SMTP)

UmanShahzad:
Contract work for edX Enterprise.

This PR adds the ability for the reporting scripts to take into account reporting configurations with arbitrary data and report types, e.g. Catalog CSV reports, Progress CSV reports, and so on.

In particular it's meant to fulfill the use case of direct integration customers who cannot actually use the API directly, to obtain their catalog data instead through CSV format by re-using the existing reporting configuration code infrastructure.

*Test Instructions*:

1. Use the Docker Devstack here. Bring up all containers.
2. Make sure you have the Enterprise feature enabled.
3. Create an Enterprise Customer. No special configuration is required.
4. Create an Enterprise Customer Catalog. The default fillings for fields should do -- just link it to the customer from before.
5. Create an Enterprise Customer Reporting Configuration. Link it to the customer, and fill in whatever fields make sense, even with fake data if needed. The data type should be 'Catalog' though, and the config should be active.
6. Get this repository, and change the `enterprise_reporting.send_enterprise_reports.should_deliver_report` function to always return `True` temporarily (so you don't have to deal with schedule issues).
7. According to the delivery method you chose in the reporting config, put a `from pdb import set_trace;set_trace()` line after `data_report_zipped = super().send(files)` is called in the appropriate `send` method (or use whatever debugging technique you know). This is because you likely don't have a proper FTP server setup for SFTP nor proper SMTP settings in your devstack to send email, so it will just crash. You want to stop abruptly here to investigate the CSV and ZIP files produced by the script before they get deleted.
8. Run this monster:
    ```
    LMS_OAUTH_KEY=discovery-key LMS_OAUTH_SECRET=discovery-secret LMS_ROOT_URL=http://localhost:18000 LMS_OAUTH_HOST=http://localhost:18000 LMS_FERNET_KEY='DUMMY KEY CHANGE BEFORE GOING TO PRODUCTION' SEND_EMAIL_FROM='my@email.com' python enterprise_reporting/send_enterprise_reports.py
    ```
9. If all goes well, and your configuration was OK, you'll reach the debugger's stop point. Go to your system's `/tmp` folder to find the appropriate ZIP file produced by the script.
10. Investigate the contents of the zip file, and ensure they make sense. This will also depend upon what you have configured in your course discovery.


edx-webhook:
Thanks for the pull request, @UmanShahzad! I've created [OSPR-2378](https://openedx.atlassian.net/browse/OSPR-2378) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

If you like, you can add yourself to the [AUTHORS](https://github.com/edx/edx-enterprise-data/blob/master/AUTHORS)      file for this repo, though that isn't required.     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

enterprise_reporting/utils.py
```diff
@@ -104,7 +104,8 @@ def is_current_time_in_schedule(frequency, hour_of_day, day_of_month=None, day_o
         elif frequency == FREQUENCY_TYPE_DAILY:
             return True
 
-    return False
+    #return False
```
UmanShahzad: Note: temporary.

enterprise_reporting/clients/__init__.py
```diff
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import os
-from time import time
+from datetime import datetime, timezone
```
pomegranited: Throws an `ImportError` for `timezone`.  But `timezone` can be removed anyway; it's not used.

enterprise_reporting/reporter.py
```diff
@@ -93,156 +52,113 @@ class EnterpriseReportSender(object):
...
+
+        LOGGER.info('Beginning to write content metadata groups to CSVs...')
+        files = []
+        for content_type, grouped_items in grouped_content_metadata.items():
+            with open(self.data_report_file_name_with.format(content_type), 'w') as data_report_file:
+                writer = csv.writer(data_report_file)
+                for item in grouped_items:
+                    writer.writerow(flatten_dict(item, target='key'))
+                    writer.writerow(flatten_dict(item, target='value'))
+                    writer.writerow([])
```
pomegranited, line 173: The CSV file resulting from these lines is not conventional -- it's writing separate heading lines for each row, and then putting an empty line in between.

Do we expect the Enterprise Customer Catalog API call to return different fields for each course?  If not, then I'd recommend just writing the `targe='key'` for the first item only, and removing the empty line.

enterprise_reporting/send_enterprise_reports.py
```diff
@@ -27,34 +28,41 @@ def send_data(config):
...
     """
     Clean up any files created by sending the enterprise report.
     """
-    directory = EnterpriseReportSender.FILE_WRITE_DIRECTORY
     pattern = r'{}'.format(enterprise_id)
-    for f in os.listdir(directory):
+    for f in os.listdir('/tmp'):
         if re.search(pattern, f):
-            os.remove(os.path.join(directory, f))
+            os.remove(os.path.join('/tmp', f))
```
pomegranited: Could you reinstate the `FILE_WRITE_DIRECTORY` constant to define `/tmp` as the destination dir?

enterprise_reporting/clients/enterprise.py
```diff
@@ -30,16 +33,44 @@ def get_all_enterprise_reporting_configs(self, **kwargs):
...
+                'enterprise_customer': enterprise_customer_uuid,
+                'page_size': 1000,
+            },
+        )
+        for catalog in enterprise_customer_catalogs.get('results', []):
+            catalog_content = self._load_data(
+                self.ENTERPRISE_CUSTOMER_CATALOGS_ENDPOINT,
+                resource_id=catalog['uuid'],
+                should_traverse_pagination=True,
+                querystring={'page_size': 1000},
```
pomegranited: Can you add paging parameters to the command-line script, to avoid having to hardcode these here?

enterprise_reporting/reporter.py
```diff
@@ -93,156 +52,113 @@ class EnterpriseReportSender(object):
...
+                decrypt_string(reporting_config['encrypted_password']),
+            )
+        elif delivery_method_str == 'sftp':
+            LOGGER.debug('{} is configured to send the report via SFTP to {}'.format(
+                enterprise_customer_name,
+                reporting_config['sftp_hostname'],
             ))
+            delivery_method = SFTPDeliveryMethod(
+                reporting_config,
+                decrypt_string(reporting_config['encrypted_sftp_password']),
```
pomegranited: Since `encrypted_sftp_password` is not required for SFTP delivery methods (unlike how [`encrypted_password` is required for Email delivery method](https://github.com/edx/edx-enterprise/pull/308/files#diff-4bcae564bc272f5c2e6753dc6021eb1cL1545)), `reporting_config['encrypted_sftp_password']` can be `None`, which causes an encoding error in `decrypt_string`.

enterprise_reporting/reporter.py
```diff
@@ -93,156 +52,113 @@ class EnterpriseReportSender(object):
...
-            os.path.join(self.file_path, os.path.basename(enterprise_report_file_name))
-        )
-        sftp.close()
-        ssh.close()
-        LOGGER.info('Successfully sent report via sftp for {}'.format(self.enterprise_customer_name))
+    def send_enterprise_report(self):
+        """Generate the report file of the appropriate type and send it through the configured delivery method."""
+        LOGGER.info('Starting process to send report to {}'.format(self.enterprise_customer_name))
+        files = self._generate_enterprise_report()
+        self.delivery_method.send(files)
```
pomegranited: This dies with a cryptic error [here](https://github.com/edx/edx-enterprise-data/pull/27/files#diff-9f1bd50e5c955841f746eadcb761f6d3R42) if no files are generated (i.e. if no courses are attached to this enterprise customer).
Could you error more gracefully here?  Stack trace:
```
ERROR:__main__:Data report failed to send for ENT-437
Traceback (most recent call last):
  File "enterprise_reporting/send_enterprise_reports.py", line 35, in send_data
    reporter.send_enterprise_report()
  File "/edx/src/edx-enterprise-data/enterprise_reporting/reporter.py", line 112, in send_enterprise_report
    self.delivery_method.send(files)
  File "/edx/src/edx-enterprise-data/enterprise_reporting/delivery_method.py", line 111, in send
    data_report_zipped = super().send(files)
  File "/edx/src/edx-enterprise-data/enterprise_reporting/delivery_method.py", line 34, in send
    data = compress_and_encrypt(files, self.password)
  File "/edx/src/edx-enterprise-data/enterprise_reporting/utils.py", line 42, in compress_and_encrypt
    zipfile = re.sub(r'_(\w+)\.(\w+)$', '.zip', files[0].name)
IndexError: list index out of range
INFO:__main__:Finished job to send report for ENT-437
```

requirements/dev.txt
```diff
@@ -49,30 +49,30 @@ kombu==3.0.37             # via celery
...
 mccabe==0.6.1             # via pylint
-mysql-connector==2.2.3
+mysql-connector==2.1.6
 packaging==17.1           # via caniusepython3
 paramiko==2.4
-path.py==11.0             # via edx-i18n-tools
-pip-tools==1.11.0
-pkginfo==1.4.1            # via twine
+path.py==11.0.1           # via edx-i18n-tools
+pip-tools==2.0.1
```
pomegranited: Could you use `pip-tools==2.0.2` [as above](https://github.com/edx/edx-enterprise-data/pull/27/files#diff-7fe11226cff646f5d9f35faa76217059R33)?  I got install compatibility errors when I hit this requirement.

requirements/quality.txt
```diff
@@ -60,21 +60,22 @@ lazy-object-proxy==1.3.1  # via astroid
...
+more-itertools==4.1.0     # via pytest
 packaging==17.1           # via caniusepython3
 paramiko==2.4
-path.py==11.0             # via edx-i18n-tools
-pbr==3.1.1                # via mock
-pip-tools==1.11.0
-pkginfo==1.4.1            # via twine
+path.py==11.0.1           # via edx-i18n-tools
+pbr==4.0.2                # via mock
+pip-tools==2.0.1
```
pomegranited: [ditto](https://github.com/edx/edx-enterprise-data/pull/27/files#diff-ca2f365d2a20e31a4a847b171f063b11R56)

pomegranited (approved):
I've left some comments and questions, but this works as described, so :+1: once these are addressed.

This repo and the added command could use some documentation too.

- [x] I tested this on my enterprise devstack using the PR instructions.
  I did *not* test the SMTP and SFTP sending of the report itself, just the generated CSV and encrypted ZIP file contents.
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike> N/A, command line only.
- [ ] Includes documentation -- docstrings only, but user docs would be useful.

enterprise_reporting/clients/__init__.py
UmanShahzad: :+1:

enterprise_reporting/clients/enterprise.py
UmanShahzad: This will require me to get a bit dirty in the code to get the page size argument passed into this function. Instead, I'll let it come from environment variables as a lot of other variables in this script are.

enterprise_reporting/reporter.py
UmanShahzad: Yeah it seems the password was never used for SFTP but [was definitely expected](https://github.com/edx/edx-enterprise-data/blob/master/enterprise_reporting/reporter.py#L54). I'm not sure if that's intentional or something missed. But in this new code, we're sending a ZIP file through SFTP whereas before it was just the plain CSV, and we put a password on it using the `encrypted_sftp_password`.

This shouldn't break any existing reporting configurations using SFTP since it seems to have already been required (but just not used), so I'll continue with it, but maybe @mattdrayer or @georgebabey know more about what the situation should be.

enterprise_reporting/reporter.py
UmanShahzad: :+1:

enterprise_reporting/reporter.py
UmanShahzad: The issue is that JSON objects with lists in them can be of varying sizes. When they vary, you'll get mismatches in columns if you arbitrarily choose a JSON object to have its schema serve as the header. The only way around this is to look through *all* JSON objects, gather all of their common keys and get their largest lists, at an arbitrarily nested level, make the header out of _that_, and then figure out how you want to write all JSON objects with `target='value'` such that they match up with a schema that may not match theirs. It's a really dirty problem that when I got into I kept hitting a million edge cases and ultimately gave up for the sake of time (and sanity).

So I discussed this solution on the upstream ticket. Any concerns @mattdrayer?

enterprise_reporting/send_enterprise_reports.py
UmanShahzad: Yes that's fine with me. I actually chose to remove it because `/tmp` is a standard directory for temporary files and less variables can help make code easier to understand sometimes. But there's no harm in bringing it back I suppose.

requirements/dev.txt
UmanShahzad: :+1:

requirements/quality.txt
UmanShahzad: :+1:

UmanShahzad:
@georgebabey This should be ready for review. How dowe test it fully on some live Jenkins server though?

Also, I'd like to ask because I feel unclear about this: what version of Python is the Jenkins job going to be running the enterprise reporting script with?

enterprise_reporting/reporter.py
mattdrayer: I agree @pomegranited -- it is not exactly what we want.  In an ideal world, yes, we would have a single header row and all of the content items populated in subsequent rows, with null/empty values for the missing fields.  However, in the spirit of rapid iteration and delivery, I think we can start with this approach and then circle back at a later time, if needed.  @georgebabey do you agree?

georgebabey: I agree as long as we let the consumers know this may change. I always worry once we release it makes it significantly more difficult to change our minds. I do wonder if having the option to send just the json (via sftp, smtp) would be more digestible if they were doing an integration based on the file...

mattdrayer:
I think the thing is that not every customer will necessarily be pushing this file into an ETL workflow -- they may simply receive the file via email and distribute the enrollment URLs to their employees using copy/paste.

enterprise_reporting/reporter.py
mattdrayer: I think the thing is that not every customer will necessarily be pushing this file into an ETL workflow -- they may simply receive the file via email and distribute the enrollment URLs to their employees using copy/paste.

georgebabey:
To test this we'll need https://github.com/edx/edx-enterprise/pull/308 to land so we can make the configuration change. Then we can test this by having jenkins point to this branch.

enterprise_reporting/reporter.py
UmanShahzad: @mattdrayer In that case JSON still shouldn't hurt right? They can search for the enrollment URL key and just copy and paste while hitting 'next' in the search. And if it's JSON they can always change their minds later to do ETL without having to concern us.

Overall I don't see the advantage of CSV over JSON, at least for the kind of data we're sending here, unless the client specifically asked for it because they have some automation around CSVs already in place, or some other policy.

On that note, if we do decide to change our minds and just send a JSON file, do let me know, I can get that done really quickly.

UmanShahzad:
@georgebabey Ok sounds good!

enterprise_reporting/reporter.py
mattdrayer: I think the issue here is that we have two business contexts jockeying for priority.  

The first is the "integration sans API" context, where the format needs to be something conventional and I agree we could send JSON and that should be easy enough for an engineer to process as part of an ETL workflow.

The second is the "HR manager of B2B customer" context, where they aren't able to perform an integration, and they aren't able to implement an ETL process -- they are going to key the information into their LMS (or simply distribute to people) manually.  For this a nice CSV file makes sense.

That all said, I think the second scenario is something that is going to require a lot more work both in terms of development and operation, so let's just toss out the CSV aspect and go with a raw JSON file for now.  If we find that customers want a nicer representation such as CSV we can tackle that as a follow-on enhancement.  Everyone good w/ that?

UmanShahzad: @mattdrayer That sounds good to me. Because we have a `report_type` variable, I will leave all of this CSV code in here so it can used at a later time when needed. I'll introduce a `json` report type, and add the code here to just insert the JSON response into a file and send that.

So if you configure customers with data type = catalog and report type = json, that should be what we want.

Will wait for @georgebabey to :+1: first though.

mattdrayer: Cool, sounds good!

UmanShahzad:
@georgebabey See https://github.com/edx/edx-enterprise-data/pull/27#discussion_r187090358 above -- if that sounds good to you, I will push commits to push JSON as a file.

enterprise_reporting/reporter.py
georgebabey: 👍 Thanks @UmanShahzad and @mattdrayer   That approach sounds good!

UmanShahzad: Great! I pushed 31bfe90af6ba7eedf1ef918d8df9863acab0ca7c.

UmanShahzad: Also made https://github.com/edx/edx-enterprise/pull/311.

UmanShahzad:
Note: I am basing the work for ENT-979 off of this, necessarily.

UmanShahzad:
Actually the commits on this PR are a lot but they look good. @georgebabey Do you want me to squash or just go with these 17? They may be informative when looking in history, but I think squashing may be policy, I can't remember.

georgebabey (approved):
Works great - manually tested through the jenkins job:
- json & csv catalogs
- multiple email recipients
- multiple active reporting configurations (progress & catalog

georgebabey:
@UmanShahzad squashing after the code-review is general best practice here.

UmanShahzad:
@georgebabey OK, sounds good :+1:

UmanShahzad:
I will likely have to run `make upgrade` before merging because tests will fail, see https://github.com/edx/edx-enterprise-data/pull/34#issuecomment-389213856. The latest version of `caniusepython3` fixes the issues now.

UmanShahzad:
Oh wow, things passed. Wut!

openedx-webhooks:
Although this pull request is already merged, I've created [OSPR-5542](https://openedx.atlassian.net/browse/OSPR-5542) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks again for your contribution.

PR #33, opened 2018-05-08
Add course key field to enterprise enrollment model

mattdrayer:
@georgebabey here is the Data API pull request to include `course_key` in the response


enterprise_reporting/reporter.py
```diff
@@ -92,6 +92,7 @@ class EnterpriseReportSender(object):
         'time_spent_hours',
         'last_activity_date',
         'user_current_enrollment_mode',
+        'course_key',
```
georgebabey: Unfortunately this won't work since Vertica wont have this field. Everything else looks good though and these changes will result in the data being in the data-api

georgebabey (changes requested):

enterprise_reporting/reporter.py
mattdrayer: Should I add a changeset to another repo in order to get the field into Vertica?  I'd need to know which one to modify.  Otherwise I could just remove this bit.

georgebabey: Just remove this line and we can roll this out for the api.

Once ENT-979 is finished we won't be using Vertica for reports anyways so no need to make changes there.

mattdrayer: Fair enough -- I've removed the line as requested.

georgebabey (approved):
Looks good! Thanks for adding this

PR #34, opened 2018-05-11
[ENT-979] Use Enterprise Data API rather than Vertica for reporting.

UmanShahzad:
*Testing Instructions*:

1. As usual, set up all the proper Enterprise-related config in your LMS, e.g. an Enterprise Customer, a reporting config, etc.. Make sure the reporting config is for CSV `progress_v2` reports.
1. Clone this repo at this branch.
1. Set up a local version of https://github.com/edx/edx-analytics-data-api.
1.1. Clone the repo and make a virtual environment for it (python 2.7).
1.2. `make develop`
1.3. `pip install -Ue /path/to/edx-enterprise-data/on/your/machine` (the `make develop` command will have already installed a different version than this branch's, so we need to override it with this)
1.4. `./manage.py migrate --noinput`
1.5. `./manage.py migrate --noinput --database=analytics`
1.6. Temporarily change `analyticsdataserver.settings.local.LMS_BASE_URL` to have the proper port of your LMS, if necessary (e.g. change from 8000 to 18000 for the docker devstack).
1.7. Temporarily change `/path/to/edx-enterprise-data/enterprise_data/fixtures/enterprise_enrollment.json` to have the `enterprise_id` for the last 2 objects match your own Enterprise Customer's UUID.
1.8. `python manage.py loaddata /path/to/edx-enterprise-data/enterprise_data/fixtures/enterprise_enrollment.json --database analytics`
1.8. `./manage.py runserver` (if your LMS is on port 8000, you need to specify the port here).
2. As in the last PR, modify `enterprise_reporting.send_enterprise_reports.should_deliver_report` to `return True` by default.
3. Put `from pdb import set_trace;set_trace()` below `data_report_zipped = super().send(files)` inside of the delivery methods so you stop short of actually delivering anything.
4. Run `LMS_OAUTH_KEY=discovery-key LMS_OAUTH_SECRET=discovery-secret LMS_ROOT_URL=http://localhost:18000 LMS_OAUTH_HOST=http://localhost:18000 LMS_FERNET_KEY='DUMMY KEY CHANGE BEFORE GOING TO PRODUCTION' SEND_EMAIL_FROM='uman@mslm.me' ANALYTICS_API_URL=http://localhost:8000 python enterprise_reporting/send_enterprise_reports.py`
5. Check the `/tmp` directory for the zip file, open it up with the appropriate password, and ensure the columns and rows look good.

PR #36, opened 2018-05-25
ENT-979. Make enterprise_id UUID for correct filtering

georgebabey:
As a string, only exact matches would properly filter. This was an issue when passing in a hyphenated UUID because the ids in the result store are just the 32 hex digits, no hyphens.


georgebabey:
@UmanShahzad This is preventing the testing of ENT-979. See https://github.com/edx/edx-enterprise-data/pull/34/files#diff-ecf5d037d5373c24a5fb31678ddf6bffL141 for where we were converting to raw hex when looking up in Vertica. Rather than do that in the consumer, this change makes it more flexible on the API side to accept anything that can be a UUID.

enterprise_data/migrations/0005_auto_20180524_2204.py
```diff
@@ -0,0 +1,20 @@
...
+# Generated by Django 1.11.5 on 2018-05-24 22:04
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('enterprise_data', '0004_auto_20180508_1652'),
```
UmanShahzad, line 11: There is apparently some issue with finding this node in the tests:

```
django.db.migrations.exceptions.NodeNotFoundError: Migration enterprise_data.0005_auto_20180524_2204 dependencies reference nonexistent parent node ('enterprise_data', '0004_auto_20180508_1652')
```

Although I'm not sure why that is...

UmanShahzad:
@georgebabey There may be some issue with travis caching or something. I'm going to try to re-open the PR and also try clearing the cache (if that's even possible).

UmanShahzad:
@georgebabey I've totally cleared the TravisCI cache here in hopes that it'll fix the above issue.

UmanShahzad:
@georgebabey I had to force re-push your branch after rebasing on master to get TravisCI to pick up some new changes. You may have to checkout this branch afresh locally.

UmanShahzad (approved):
:+1: Looks good.

PR #42, opened 2018-06-21
Adds an endpoint for dashboard analytics

adamstankiewicz:
For the enterprise admin dashboard project, we would like to display aggregate data about an enterprise's learners. The current endpoint in the enterprise data api returns a detailed list of all course enrollments for an enterprise but does not offer any insights into aggregate data. 

[ENT-1056](https://openedx.atlassian.net/browse/ENT-1056) describes 3 fields we would like to see in the admin dashboard for now. This PR creates a new endpoint that returns those 3 fields:
1. Number of enrolled learners;
2. Number of active learners in the:
    - Past week
    - Past month
3. Number of course completions.

Example response:
```json
{
    "active_learners": {
        "past_month": 2,
        "past_week": 1
    },
    "enrolled_learners": 3,
    "course_completions": 2
}
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,10 @@ Change Log
 Unreleased
 ----------
 
+[0.1.8] - 2018-06-29
+--------------------
+* Introduce endpoint for returning summary data about enterprise enrollments.
+
 [0.1.7] - 2018-06-28
 --------------------
 * Make the enterprise enrollment schema match the field changes made in the pipeline.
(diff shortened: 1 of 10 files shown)


brittneyexline (commented):
I think this looks good so far, but I'm wondering if it's worth making an EnterpriseEnrollmentsViewSet so that all of the serializer, filter, authentication, etc. stuff doesn't have to get redefined, and then this new endpoint would just be a detail route on the ViewSet. Of course, that means touching the original endpoint...

georgebabey:
Good call @brittneyexline  A ViewSet would make sense here - I'm especially concerned if we don't someone may forget the consentGranted filter or the enterprise permission class. I suspect we'll keep adding additional endpoints so lets tackle that now.

brittneyexline (approved):

georgebabey (approved):

PR #43, opened 2018-06-28
ENT-992 Modify analytics table to add/modify fields to match

brittneyexline:


enterprise_data/models.py
```diff
@@ -61,7 +61,8 @@ class Meta:
     last_activity_date = models.DateField(null=True)
     coupon_name = models.CharField(max_length=128, null=True)
     coupon_code = models.CharField(max_length=128, null=True)
-    final_grade = models.FloatField(null=True)
+    offer = models.CharField(max_length=128, null=True)
+    current_grade = models.FloatField(null=True)
```
adamstankiewicz, line 65: @brittneyexline Just a small clarification question: Is the `current_grade` field non-null if the user hasn't passed/completed the course? The original `final_grade` name makes it seem like this field is only non-null if the user has passed/completed. `current_grade` does make more sense if it's continually updated as they progress through the course.

brittneyexline: I renamed it because final_grade was misleading - it is actually the current grade value in the grades table. I believe it could still be null if the user has enrolled in the course but hasn't completed any graded assignments.

adamstankiewicz (approved):

PR #44, opened 2018-07-10
Refactor consent filter backend & support sorting for enrollments endpoint

adamstankiewicz:
1. Remove the need to override the `filter_queryset` in `EnterpriseViewSet`;
2. Update the `enrollments` enrollment to support sorting using the built-in ordering from `rest_framework`.


georgebabey (approved):
Change looks good, although I'm surprised/sad theres no tests for the view 😿

PR #45, opened 2018-07-20
Return number of enterprise users in overview endpoint

jonalexvasquez:
Added functionality so that number of users(enrolled AND not enrolled) is returned in the overview endpoint. Changed the use of user id from "enterprise_user_id" to "lms_user_id" based on what was discussed at standup

### DIFF
--- a/enterprise_data/api/v0/views.py
+++ b/enterprise_data/api/v0/views.py
@@ -18,7 +18,7 @@ from django.db.models import Max
 
 from enterprise_data.api.v0 import serializers
 from enterprise_data.filters import ConsentGrantedFilterBackend
-from enterprise_data.models import EnterpriseEnrollment
+from enterprise_data.models import EnterpriseEnrollment, EnterpriseUser
 from enterprise_data.permissions import IsStaffOrEnterpriseUser
 
 LOGGER = getLogger(__name__)
@@ -98,6 +98,13 @@ class EnterpriseEnrollmentsViewSet(EnterpriseViewSet, viewsets.ModelViewSet):
         """
         return queryset.filter(has_passed=1)
 
+    def filter_number_of_users(self):
+        """
+        Returns number of enterprise users (enrolled AND not enrolled learners)
+        """
+        enterprise_id = self.kwargs['enterprise_id']
+        return EnterpriseUser.objects.filter(enterprise_id=enterprise_id)
+
     def subtract_one_month(self, original_date):
         """
         Returns a date exactly one month prior to the passed in date.
@@ -130,6 +137,7 @@ class EnterpriseEnrollmentsViewSet(EnterpriseViewSet, viewsets.ModelViewSet):
             - # of enrolled learners;
             - # of active learners in the past week/month;
             - # of course completions.
+            - # of enterprise users (LMS)
         """
         enrollments = self.get_queryset()
         enrollments = self.filter_queryset(enrollments)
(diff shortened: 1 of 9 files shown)


enterprise_data/models.py
```diff
@@ -35,7 +35,6 @@ class Meta:
 
     enterprise_id = models.UUIDField()
     enterprise_name = models.CharField(max_length=255)
-    lms_user_id = models.PositiveIntegerField()
```
brittneyexline: we shouldn't modify the enrollment data model - we need to keep the table as is.

enterprise_data/models.py
```diff
@@ -80,3 +75,41 @@ def __repr__(self):
         Return uniquely identifying string representation.
         """
         return self.__str__()
+
+@python_2_unicode_compatible
+class EnterpriseUsers(models.Model):
```
brittneyexline: rename this to EnterpriseUser - generally model class names are singular

enterprise_data/models.py
```diff
@@ -80,3 +75,41 @@ def __repr__(self):
...
+        return "<Enterprise User {user} in {enterprise}>".format(
+            user=self.lms_user_id,
+            enterprise=self.enterprise_id
+        )
+
+    def __repr__(self):
+        """
+        Return uniquely identifying string representation.
+        """
+        return self.__str__()
```
brittneyexline: add a new line here

enterprise_data/api/v0/views.py
```diff
@@ -96,6 +96,12 @@ def filter_course_completions(self, queryset):
         """
         return queryset.filter(has_passed=1)
 
+    def filter_number_of_users(self):
+        """
+        Returns number of enterprise users (enrolled AND not enrolled learners)
+        """
+        return EnterpriseUsers.objects.filter(enterprise_id=enterprise_id)
```
brittneyexline: where does the variable enterprise_id come from? in the other queryset, it looks like they get it from the request kwargs: `enterprise_id = self.kwargs['enterprise_id']`

brittneyexline (commented):
good start here, just a few small things, and also tests :-)

enterprise_data/api/v0/views.py
```diff
@@ -96,6 +96,12 @@ def filter_course_completions(self, queryset):
         """
         return queryset.filter(has_passed=1)
 
+    def filter_number_of_users(self, queryset):
+        """
+        Returns number of enterprise users (enrolled AND not enrolled learners)
+        """
+        return queryset.values_list('enterprise_id', flat=True).distinct()
```
brittneyexline: shouldn't we be using the EnterpriseUser model here to determine this? I thought you were doing that previously when I reviewed this (https://github.com/edx/edx-enterprise-data/pull/45/commits/7091417c47845ff753aded612b1a4ad9a50959d9#diff-e50e61ec499ff572ac7766646ddaa2d1R103)

brittneyexline (commented):
Outside of the one review comment, be sure to rebase and squash your commits, and also update the package version and changelog like so: https://github.com/edx/edx-enterprise-data/pull/50

enterprise_data/api/v0/views.py
jonalexvasquez: You're right! I have no idea what I was thinking...

brittneyexline (approved):
good to go, just need to update the version before merging

enterprise_data/api/v0/views.py
```diff
@@ -139,15 +147,14 @@ def overview(self, request, **kwargs):  # pylint: disable=unused-argument
         past_month_date = self.subtract_one_month(date.today())
         active_learners_week = self.filter_active_learners(enrollments, past_week_date)
         active_learners_month = self.filter_active_learners(enrollments, past_month_date)
-        last_updated_date = self.get_max_created_date(enrollments)
```
brittneyexline: we need to keep this last_updated_date logic - looks like you accidentally deleted it here and below.

enterprise_data/migrations/0001_initial.py
```diff
@@ -46,4 +46,24 @@ class Migration(migrations.Migration):
                 'verbose_name_plural': 'Enterprise Enrollments',
             },
         ),
+        migrations.CreateModel(
```
brittneyexline, line 49: instead of modifying this migration, we need to create a new migration. you will have to remove this code and run ./manage.py makemigrations in your edx-enterprise-data directory

PR #46, opened 2018-07-31
ENT-1126 Gate access to admin users based on group membership

brittneyexline:
Similar to how we gate access to the enrollment api by checking membership in a django group, this PR adds the check to the data api to restrict access only to users we have granted it to.


georgebabey (approved):
Looks good!

PR #58, opened 2018-08-29
ENT-1164 | Adding backend support for gettings EnterpriseUser data

christopappas:
ENT-1164 | Adding backend support for gettings EnterpriseUser data.

Includes new viewset, serializer, and an updated ForeignKey relationship between EnterpriseEnrollment and EnterpriseUser models. As such, a migration is included in this PR.

Additionally, bumped up the version of tox-battery from 0.5 to 0.5.1 in some requirements files, as they caused local development pip issues.


enterprise_data/api/v0/views.py
```diff
@@ -156,3 +155,24 @@ def overview(self, request, **kwargs):  # pylint: disable=unused-argument
...
+    serializer_class = serializers.EnterpriseUserSerializer
+
+    def get_queryset(self):
+        """
+        Returns learner records for a given enterprise.
+        """
+        enterprise_id = self.kwargs['enterprise_id']
+        users = EnterpriseUser.objects.filter(enterprise_id=enterprise_id)
+
+        has_enrollments = self.request.query_params.get('has_enrollments')
```
brittneyexline: i think this filter logic is specific to the list endpoint on the viewset... i'm not sure it would make sense for all endpoints on the viewset, current or future. i would move this has_enrollments logic to the list endpoint specifically by overriding the list method.

enterprise_data/models.py
```diff
@@ -34,7 +34,11 @@ class Meta:
     enterprise_id = models.UUIDField()
     enterprise_name = models.CharField(max_length=255)
     lms_user_id = models.PositiveIntegerField()
-    enterprise_user_id = models.PositiveIntegerField()
+    enterprise_user = models.ForeignKey(
```
brittneyexline, line 37: this makes sense, but we should check that we are inserting user data before enrollment data, otherwise our data pipeline job for enrollments might break.

enterprise_data/api/v0/serializers.py
```diff
@@ -24,3 +24,13 @@ def get_course_api_url(self, obj):
...
+
+
+class EnterpriseUserSerializer(serializers.ModelSerializer):
+    """
+    Serializer for EnterpriseUser model.
+    """
+
+    class Meta:
+        model = EnterpriseUser
+        fields = '__all__'
```
brittneyexline: not sure if we should also exclude 'created' here? 'created' is when the pipeline inserted the data.

brittneyexline (commented):
looks good overall, left a few comments for things to consider.

enterprise_data/api/v0/views.py
christopappas: ah, good point. will do

enterprise_data/api/v0/views.py
```diff
@@ -156,3 +155,32 @@ def overview(self, request, **kwargs):  # pylint: disable=unused-argument
...
+        List view for learner records for a given enterprise.
+        """
+        enterprise_id = kwargs['enterprise_id']
+        users = self.queryset.filter(enterprise_id=enterprise_id)
+
+        has_enrollments = request.query_params.get('has_enrollments')
+        if has_enrollments == 'true':
+            users = users.filter(enrollments__isnull=False).distinct()
+        elif has_enrollments == 'false':
+            users = users.filter(enrollments__isnull=True)
```
brittneyexline, line 178: why not distinct() here as well?

brittneyexline (approved):

enterprise_data/api/v0/views.py
christopappas: `__isnull=False` results in duplicate objects returned, whereas `__isnull=True` does not. Based on my googling, it seems like a django quirk

PR #62, opened 2018-09-12
ENT-1185 - Update Course Catalog CSV flat file to have only one single header and a line of rows.

asadiqbal08:
[ENT-1185](https://openedx.atlassian.net/browse/ENT-1185)

Update Course Catalog CSV flat file format to have only one single header and a line of rows.

#### NOTE: 
Implemented a suggested solution of putting all of the values as is into the column, even for simple arrays, in the form of JSON objects. This would also likely be the easiest to implement. In the example the column would have the possible value of:

> transcript_languages

`
 ['English', 'Hindi', 'Dutch', 'Chinese - Mandarin']  
`

### DIFF
--- a/enterprise_reporting/reporter.py
+++ b/enterprise_reporting/reporter.py
@@ -15,7 +15,7 @@ from uuid import UUID
 from enterprise_reporting.clients.enterprise import EnterpriseAPIClient, EnterpriseDataApiClient
 from enterprise_reporting.clients.vertica import VerticaClient
 from enterprise_reporting.delivery_method import SFTPDeliveryMethod, SMTPDeliveryMethod
-from enterprise_reporting.utils import decrypt_string, flatten_dict
+from enterprise_reporting.utils import decrypt_string, generate_data
 
 LOGGER = logging.getLogger(__name__)
 NOW = datetime.datetime.now().strftime("%Y-%m-%d")
@@ -180,10 +180,13 @@ class EnterpriseReportSender(object):
         for content_type, grouped_items in grouped_content_metadata.items():
             with open(self.data_report_file_name_with.format(content_type), 'w') as data_report_file:
                 writer = csv.writer(data_report_file)
+                if grouped_items:
+                    # Write single row of headers in csv.
+                    writer.writerow(generate_data(grouped_items[0], target='key'))
+
                 for item in grouped_items:
-                    writer.writerow(flatten_dict(item, target='key'))
-                    writer.writerow(flatten_dict(item, target='value'))
-                    writer.writerow([])
+                    writer.writerow(generate_data(item, target='value'))
+
                 files.append(data_report_file)
         return files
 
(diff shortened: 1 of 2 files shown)


enterprise_reporting/reporter.py
```diff
@@ -180,8 +180,10 @@ def _generate_enterprise_report_catalog_csv(self):
         for content_type, grouped_items in grouped_content_metadata.items():
             with open(self.data_report_file_name_with.format(content_type), 'w') as data_report_file:
                 writer = csv.writer(data_report_file)
+                if grouped_items:
+                    writer.writerow(flatten_dict(grouped_items[0], target='key'))
```
zubair-arbi: Can you please add a comment mentioning that we are doing this to add column headers for one time only?

zubair-arbi (approved):
Just one small suggestion otherwise change looks good 👍

georgebabey:
I don't believe this will work if the first course does not have some fields, but other courses do. For example course_runs for 8647cd55-7fea-4b29-b4de-e014fd4f030a has multiple seat_types, transcript_languages,etc which were being serialized to 1 per column. This leads to inconsistent # of columns, which is why the odd format was chosen in the first place.

screen shot 2018-09-12 at 9 53 43 am

georgebabey (changes requested):
Discussion around formatting in the JIRA, but I believe this will need to be updated

saleem-latif (approved):
Looks Good 👍

enterprise_reporting/reporter.py
```diff
@@ -180,9 +180,12 @@ def _generate_enterprise_report_catalog_csv(self):
...
                 writer = csv.writer(data_report_file)
+                if grouped_items:
+                    # Write single row of headers in csv.
+                    writer.writerow(generate_data(grouped_items[0], target='key'))
+
                 for item in grouped_items:
-                    writer.writerow(flatten_dict(item, target='key'))
-                    writer.writerow(flatten_dict(item, target='value'))
+                    writer.writerow(generate_data(item, target='value'))
                     writer.writerow([])
```
georgebabey: This is adding an extra line between each row

georgebabey:
This looks good, just a couple suggestions:
- For empty lists we should probably just output an empty string rather than filling it up with []
- If only one item in a list, output that item instead of the list. (Example: Language is usually just set to one language even though it supports multiple)

asadiqbal08:
@georgebabey, I looked around your suggestions. Good ones and I am OK with 
--> "For empty lists we should probably just output an empty string rather than filling it up with []" 
But I think we should keep the data as a list even if there is one or more items in list. 

I am just thinking about the consistency b/w data representation in CSV files . 

Some **transcript_languages** will be just LIST while some other rows have strings OR the possibility of having the single item in the LIST is another DataType e.g. dict etc. 

Suggest ?? OR as you like, I will update the PR accordingly.

enterprise_reporting/reporter.py
asadiqbal08: Ok. I will remove it.

georgebabey:
@asadiqbal08 Yeah, consistency makes sense then. Lets keep it as a list

georgebabey (approved):
Just remove the extra rows and this is good to go 👍

PR #64, opened 2018-09-13
Learner activity details

mushtaqak:
This PR adds some filters for Learner Activity cards. They include:
1. Active learners in past week.
2. Inactive learners in past week.
3. Inactive learners in past month.

The criterion used to determine learner activity is `last_activity_date` field.

This PR also does some refactoring.

[ENT-1166](https://openedx.atlassian.net/browse/ENT-1166)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,8 +14,9 @@ Change Log
 Unreleased
 ----------
 
-[0.2.9] - 2018-09-19
+[0.2.9] - 2018-09-24
 --------------------
+* Update the course catalog CSV flat file to have only one single header and a line of rows in JSON form.
 * Adding filters for Learner Activity cards. These include:
     - Active learners in past week.
     - Inactive learners in past week.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,13 @@ Change Log
 Unreleased
 ----------
 
+[0.2.9] - 2018-09-19
+--------------------
+* Adding filters for Learner Activity cards. These include:
+    - Active learners in past week.
+    - Inactive learners in past week.
+    - Inactive learners in past month
+
 [0.2.8] - 2018-09-12
 --------------------
 * Adding query params on /users/ enpoint for active_courses and enrollment_count
--- a/enterprise_data/api/v0/views.py
+++ b/enterprise_data/api/v0/views.py
@@ -25,6 +25,17 @@ from enterprise_data.permissions import HasDataAPIDjangoGroupAccess
 LOGGER = getLogger(__name__)
 
 
+def subtract_one_month(original_date):
+    """
+    Returns a date exactly one month prior to the passed in date.
+    """
+    one_day = timedelta(days=1)
+    one_month_earlier = original_date - one_day
+    while one_month_earlier.month == original_date.month or one_month_earlier.day > original_date.day:
+        one_month_earlier -= one_day
+    return one_month_earlier
+
+
 class EnterpriseViewSet(viewsets.ViewSet):
     """
     Base class for all Enterprise view sets.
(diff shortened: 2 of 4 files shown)


enterprise_data/api/v0/views.py
```diff
@@ -25,6 +25,17 @@
 LOGGER = getLogger(__name__)
 
 
+def subtract_one_month(original_date):
+    """
+    Returns a date exactly one month prior to the passed in date.
+    """
+    one_day = timedelta(days=1)
+    one_month_earlier = original_date - one_day
+    while one_month_earlier.month == original_date.month or one_month_earlier.day > original_date.day:
```
zubair-arbi, line 34: Maybe we can shorten this logic and make it more readable: https://stackoverflow.com/questions/9724906/python-date-of-the-previous-month

enterprise_data/api/v0/views.py
```diff
@@ -94,6 +96,29 @@ def get_queryset(self):
...
         return enrollments
 
+    def apply_filters(self, queryset):
+        """
+        Filters enrollments based on query params.
+        """
+        query_filters = self.request.query_params
+
+        past_week_date = date.today() - timedelta(weeks=1)
+        past_month_date = subtract_one_month(date.today())
```
zubair-arbi, line 106: Maybe also name the method `subtract_one_month` to `past_month_date` for clarity

enterprise_data/api/v0/views.py
mushtaqak: This was written before. I just moved it outside the ViewSet.

enterprise_data/api/v0/views.py
mushtaqak: Ok.

enterprise_data/api/v0/views.py
```diff
@@ -104,33 +129,45 @@ def filter_active_learners(self, queryset, last_activity_date):
         """
         Filters queryset to include enrollments more recent than the specified `last_activity_date`.
         """
-        return self.filter_distinct_learners(
-            queryset.filter(last_activity_date__gte=last_activity_date)
-        )
+        return queryset.filter(last_activity_date__gte=last_activity_date)
```
georgebabey, line 132: Without the distinct here, if an active learner has multiple enrollments won't we get multiple results?

enterprise_data/api/v0/views.py
```diff
@@ -104,33 +129,45 @@ def filter_active_learners(self, queryset, last_activity_date):
...
+
+    def filter_inactive_learners(self, queryset, last_activity_date):
+        """
+        Filters queryset to include enrollments more older than the specified `last_activity_date`.
+        """
+        return queryset.filter(last_activity_date__lte=last_activity_date)
```
georgebabey, line 138: I believe we also want `filter_distinct_learners` here as well

enterprise_data/api/v0/views.py
```diff
@@ -104,33 +129,45 @@ def filter_active_learners(self, queryset, last_activity_date):
...
         """
         return queryset.filter(has_passed=1)
 
+    def filter_past_week_completions(self, queryset):
+        """
+        Filters only those learners who completed a course in last week.
+        """
+        date_today = date.today()
+        date_week_before = date_today - timedelta(weeks=1)
+        return queryset.filter(
```
georgebabey, line 158: And distinct here..

georgebabey (changes requested):
Overall looks good, just have some questions regarding using the `filter_distinct_learners` in additional places to avoid multiple rows for multiple enrollments.

enterprise_data/api/v0/views.py
mushtaqak: Actually the ticket says:
> Current columns filtered by learners active in past week
There may be duplicate learners if they were in active than more than 1 course

enterprise_data/api/v0/views.py
mushtaqak: `filter_distinct_learners` actually modifies the queryset since it only returns `enterprise_user_id` due to `queryset.values_list('enterprise_user_id', flat=True).distinct()`

enterprise_data/api/v0/views.py
georgebabey: Ah, got it, so its actually returning enrollment rows, not learners. got it. 👍

georgebabey (approved):
Sorry about the confusion on distinct. Rereading the ticket you are correct, the intent is to return enrollment records, not learners, so distinct is not needed.

:shipit:

enterprise_data/api/v0/views.py
mushtaqak: @zubair-arbi I think I will pass on this one as well. Since, in test we may need to get a date for two or three months back or in future we can set a differnt timeline for user activity as well. Let me know if you want to change this as well.

CHANGELOG.rst
```diff
@@ -14,6 +14,14 @@ Change Log
 Unreleased
 ----------
 
+[0.2.9] - 2018-09-24
+--------------------
+* Update the course catalog CSV flat file to have only one single header and a line of rows in JSON form.
```
mushtaqak, line 19: cc: @asadiqbal08

PR #66, opened 2018-09-26
ENT-1189 Add enterprise learner course completion count

zubair-arbi:
[ENT-1189](https://openedx.atlassian.net/browse/ENT-1189)

* Allow for extra field `course_completion_count` to be added to response when "extra_fields" query param is provided that includes the text "`course_completion_count`".
* Allow a new query param `all_enrollments_passed` to filter out enterprise learners on the basis of all enrollments passed or not.
* Consider `extra_fields` as a list of parameters, e.g. `'extra_fields': ['enrollment_count', 'course_completion_count']`

**Expected URL for ENT-1189:**
http://localhost:18000/enterprise/api/v0/enterprise/72416e52-8c77-4860-9584-15e5b06220fb/users/?extra_fields=enrollment_count&extra_fields=course_completion_count&has_enrollments=true&active_courses=false&all_enrollments_passed=true&page_size=50

### DIFF
--- a/enterprise_data/api/v0/serializers.py
+++ b/enterprise_data/api/v0/serializers.py
@@ -38,11 +38,16 @@ class EnterpriseUserSerializer(serializers.ModelSerializer):
     def to_representation(self, instance):
         representation = super(EnterpriseUserSerializer, self).to_representation(instance)
         request = self.context.get('request')
-        extra_fields = request.query_params.get('extra_fields')
+        extra_fields = request.query_params.getlist('extra_fields')
         if extra_fields is not None:
             if 'enrollment_count' in extra_fields:
                 enrollments = instance.enrollments.exclude(consent_granted=False)
                 representation['enrollment_count'] = enrollments.count()
+            if 'course_completion_count' in extra_fields:
+                representation['course_completion_count'] = instance.enrollments.exclude(
+                    consent_granted=False
+                ).filter(has_passed=True).count()
+
         return representation
 
 
--- a/enterprise_data/tests/test_views.py
+++ b/enterprise_data/tests/test_views.py
@@ -291,6 +291,7 @@ class TestEnterpriseUsersViewSet(APITestCase):
             enterprise_user=self.ent_user4,
             course_end=date_in_past,
             consent_granted=True,
+            has_passed=True,
         )
         EnterpriseEnrollmentFactory(
             enterprise_user=self.ent_user4,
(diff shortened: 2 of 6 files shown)


enterprise_data/api/v0/serializers.py
```diff
@@ -42,6 +42,9 @@ def to_representation(self, instance):
         if extra_fields is not None:
             if 'enrollment_count' in extra_fields:
                 representation['enrollment_count'] = instance.enrollments.count()
+            if 'course_completion_count' in extra_fields:
```
mushtaqak, line 46: Should we not have elif statement since `extra_fields` key can only contain single value.

mushtaqak (approved):
One question otherwise looks good.

enterprise_data/tests/test_views.py
```diff
@@ -433,6 +434,42 @@ def test_viewset_enrollment_count_not_present(self):
...
+        """
+        EnterpriseUserViewset should ultimately return a response that includes
+        the `course_completion_count` field if "course_completion_count" is
+        specified in the "extra_fields" query parameter value.
+        """
+        kwargs = {
+            'enterprise_id': 'ee5e6b3a-069a-4947-bb8d-d2dbc323396c',
+            'pk': self.ent_user4.id,
+        }
+        params = {'extra_fields': 'course_completion_count', }
```
mushtaqak: It's a bit confusing we are expecting a list in `to_representation` but here we can only pass single value.

mushtaqak: Also can we have a test where both `extra_fields` are used? and what will be our expectations, whom would we prioritize ?

mushtaqak (changes requested):

enterprise_data/api/v0/serializers.py
zubair-arbi: According to my understanding as name says `extra_fields` can contain multiple fields so the value can be `enrollment_count+course_completion_count`.

enterprise_data/tests/test_views.py
zubair-arbi: I will add a test for multiple fields.
We don't need to prioritize, all specified fields in `extra_fields` will be returned in the response.

enterprise_data/api/v0/serializers.py
mushtaqak: According to the feedback on https://github.com/edx/edx-portal/pull/79. we will be getting `'extra_fields'` list as follow:
```python
request.query_params.getlist('extra_fields')
```

zubair-arbi:
@mushtaqak I have addressed your feedback. Please have another look.

mushtaqak (approved):
Looks good 👍

zubair-arbi:
@adamstankiewicz @georgebabey please review so that I can merge this backend change.

zubair-arbi:
We need this PR to merge first for enterprise portal.

enterprise_data/tests/test_views.py
```diff
@@ -463,6 +514,44 @@ def test_viewset_enrollment_count_not_present(self):
         response = self.client.get(url,)
         assert 'enrollment_count' not in response.json()
 
+    def test_viewset_enrollment_count_and_course_completion_count_present(self):
+        """
+        EnterpriseUserViewset should ultimately return a response that includes
+        both the `enrollment_count` field and `course_completion_count` field
+        if value "enrollment_count+course_completion_count" is specified in
```
adamstankiewicz: Might want to update this comment as the `extra_fields` value is no longer `enrollment_count+course_completion_count`.

adamstankiewicz (approved):
Looks good. Just one small update in one of the test comments. After that, should be good to go!

PR #70, opened 2018-10-12
ENT-1235 Add filtering on audit enrollments for reporting

zubair-arbi:
Enable audit data reporting flag `enable_audit_enrollment` of enterprise customers to also be recognized by the Learner Progress reporting workflow.

- Show/Hide audit enrollment in CSV reports depending on enterprise customers data reporting flag `enable_audit_enrollment`
- Use `edx-django-utils` for caching the enterprise reporting configuration for 6 hours.
```from edx_django_utils.cache import TieredCache```

**Remaining Tasks:**
- Tests coverage

**Story History and Details:**
Currently the "Enable audit data reporting" flag is only recognized by the SAP SuccessFactors reporting workflow. The flag needs to also be recognized by the Learner Progress reporting workflow. See ENT‌-1210 for more information.

1. Should not show audits in CSV reports
2. Admin Dashboard: Filter out audit learner for both overview and UI table (just enrollment data)
    1. Audit Enrollment is filtered out
    2. Users with audit enrollment included in user count but not enrolled count
3. Enterprise Customer model (Flag is in django admin)
4. Data API cache enterprise configuration for 6 hours (for reporting)

**Example:**

- Learners who are only enrolled in audit courses will not be included as “enrolled” learners.
- Count of courses a Learner is enrolled in: If enrolled in 3 audit courses - the count for courses enrolled is 0

### DIFF
--- a/enterprise_data/clients.py
+++ b/enterprise_data/clients.py
@@ -5,12 +5,16 @@ from __future__ import absolute_import, unicode_literals
 
 import logging
 
+from edx_django_utils.cache import TieredCache
 from edx_rest_api_client.client import EdxRestApiClient
 from edx_rest_api_client.exceptions import HttpClientError, HttpServerError
 from rest_framework.exceptions import NotFound, ParseError
 
 from django.conf import settings
 
+from enterprise_data.utils import get_cache_key
+
+DEFAULT_REPORTING_CACHE_TIMEOUT = 60 * 60 * 6  # 6 hours (Value is in seconds)
 LOGGER = logging.getLogger('enterprise_data')
 
 
@@ -62,6 +66,15 @@ class EnterpriseApiClient(EdxRestApiClient):
         """
         Get the enterprises that this user has access to for the data api permission django group.
         """
+        cache_key = get_cache_key(
+            resource='enterprise-customer',
+            user=user.username,
+            enterprise_customer=enterprise_id,
+        )
+        cached_response = TieredCache.get_cached_response(cache_key)
+        if cached_response.is_found:
+            return cached_response.value
+
         try:
             querystring = {
                 'permissions': [self.ENTERPRISE_DATA_API_GROUP],
@@ -87,4 +100,5 @@ class EnterpriseApiClient(EdxRestApiClient):
         if response['count'] == 0:
             return None
 
+        TieredCache.set_all_tiers(cache_key, response['results'][0], DEFAULT_REPORTING_CACHE_TIMEOUT)
         return response['results'][0]
--- a/enterprise_data/filters.py
+++ b/enterprise_data/filters.py
@@ -27,3 +27,26 @@ class ConsentGrantedFilterBackend(filters.BaseFilterBackend):
         """
         filter_kwargs = {view.CONSENT_GRANTED_FILTER: True}
         return queryset.filter(**filter_kwargs)
+
+
+class AuditEnrollmentsFilterBackend(filters.BaseFilterBackend):
+    """
+    Filter backend to exclude enrollments where enrollment mode is `audit`.
+
+    This requires that `ENROLLMENT_MODE_FILTER` be set in the view as a class
+    variable, to identify the object's relationship to the
+    `user_current_enrollment_mode` field.
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        """
+        Filter out queryset for results where enrollment mode is `audit`.
+        """
+        enterprise_id = view.kwargs['enterprise_id']
+        enable_audit_enrollment = request.session['enable_audit_enrollment'].get(enterprise_id, False)
+
+        if not enable_audit_enrollment:
+            filter_kwargs = {view.ENROLLMENT_MODE_FILTER: 'audit'}
+            queryset = queryset.exclude(**filter_kwargs)
+
+        return queryset
--- a/enterprise_data/permissions.py
+++ b/enterprise_data/permissions.py
@@ -8,6 +8,7 @@ from logging import getLogger
 from rest_framework import permissions
 
 from enterprise_data.clients import EnterpriseApiClient
+from enterprise_data.utils import update_session_with_enterprise_data
 
 LOGGER = getLogger(__name__)
 
@@ -94,12 +95,19 @@ class HasDataAPIDjangoGroupAccess(permissions.BasePermission):
 
         if ('enterprises_with_access' not in request.session or
                 enterprise_in_url not in request.session['enterprises_with_access']):
-            request.session['enterprises_with_access'] = {}
             enterprise_data = self.get_enterprise_with_access_to(request.auth, request.user, enterprise_in_url)
             if not enterprise_data:
-                request.session['enterprises_with_access'][enterprise_in_url] = False
+                update_session_with_enterprise_data(
+                    request, enterprise_in_url, enterprises_with_access=False, enable_audit_enrollment=False
+                )
             else:
-                request.session['enterprises_with_access'][enterprise_in_url] = True
+                enable_audit_enrollment = enterprise_data.get('enable_audit_enrollment', False)
+                update_session_with_enterprise_data(
+                    request,
+                    enterprise_in_url,
+                    enterprises_with_access=True,
+                    enable_audit_enrollment=enable_audit_enrollment
+                )
 
         permitted = request.session['enterprises_with_access'][enterprise_in_url]
         if not permitted:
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

**Comment by zubair-arbi** [comment]
This PR is ready for initial review.

**Comment by christopappas** [comment]
@zubair-arbi looks good from what I can see. I wanted to double check though, there is an `enrollment_count` optionally serialized field in the `EnterpriseUser` serializer (https://github.com/edx/edx-enterprise-data/blob/0136418f4626d83d510d5055e979d086a6ee7512/enterprise_data/api/v0/serializers.py#L38).

Should this also filter auditing users?

File: enterprise_reporting/clients/enterprise.py [hunk 1]
```diff
@@ -8,7 +8,13 @@
 import os
 from collections import OrderedDict
 
+from edx_django_utils.cache import TieredCache
+
 from enterprise_reporting.clients import EdxOAuth2APIClient
+from enterprise_reporting.utils import get_cache_key
+
+
+DEFAULT_REPORTING_CACHE_TIMEOUT = 3600 * 60 * 6  # 6 hours (Value is in seconds)
```
brittneyexline: is the math here right? i think it should be 60 * 60 * 6 for 6 hours (in seconds)

brittneyexline (commented):
I think this is a good start, but my understanding of where we would do the filtering is different than what is here. I believe we wanted to filter at the data api layer so that the admin dashboard would also automatically get the enrollment data filtered appropriately, and then so would the reporting job. Here, it seems to be implemented in the reporting job only. I would move most of what you've implemented for the client/caching to the enterprise_data folder where the data api code lives.

zubair-arbi:
> @zubair-arbi looks good from what I can see. I wanted to double check though, there is an `enrollment_count` optionally serialized field in the `EnterpriseUser` serializer (
> 
> [edx-enterprise-data/enterprise_data/api/v0/serializers.py](https://github.com/edx/edx-enterprise-data/blob/0136418f4626d83d510d5055e979d086a6ee7512/enterprise_data/api/v0/serializers.py#L38)
> 
> Line 38 in [0136418](/edx/edx-enterprise-data/commit/0136418f4626d83d510d5055e979d086a6ee7512)
>  def to_representation(self, instance): 
> ).
> 
> Should this also filter auditing users?

You are right, in ticket its mentioned that "`Users with audit enrollment included in user count but not enrolled count`". I will update it.

zubair-arbi:
> I think this is a good start, but my understanding of where we would do the filtering is different than what is here. I believe we wanted to filter at the data api layer so that the admin dashboard would also automatically get the enrollment data filtered appropriately, and then so would the reporting job. Here, it seems to be implemented in the reporting job only. I would move most of what you've implemented for the client/caching to the enterprise_data folder where the data api code lives.

So we need to implement filtering in the pipeline `edx-analytics-pipeline`?

zubair-arbi:
@christopappas @brittneyexline I have addressed your feedback. Just need clarification regarding if I implemented the filtering on right place, which according to my understanding we needed to add in views.

brittneyexline:
@zubair-arbi - We do need the filtering in the data api, but the fetch to get the data that determines if we should filter should also be done in the data api and not as a flag on the endpoint, so that it happens automatically for any caller of the api. We do not need to make any changes in the pipeline.
I would imagine a new filter class here: https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/filters.py
This filter would call the enterprise api as you currently have it in the enterprise_reporting/ folder. It would be good to consolidate the caching of the call with the call we make for determining permissions here: https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/permissions.py#L83
As such, I would expect the caching logic to be applied here: https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/clients.py
Keep in mind that enterprise_data/ is what gets installed in the analytics data api and serves up the enterprise data api, and enterprise_reporting/ is what gets run by a jenkins job that calls that api. We want the logic for determining if we should filter and performing the filtering to all live in enterprise_data/, so that all callers of the api will get the appropriate filtering without having to know if they are configured to filter out audit enrollments or not.
Let me know if you have any other questions!

christopappas:
@zubair-arbi -- as for my part of the feedback, I think it looks good

I would just ask that you write a test (https://github.com/edx/edx-enterprise-data/blob/29de10e5a0c6feab79c1c929fd018665783fe3c2/enterprise_data/tests/test_views.py#L259) to prove everything looks good for the `enrollment_count` value being returned. Maybe adding another

```
        EnterpriseEnrollmentFactory(
            enterprise_user=self.ent_user7,
            consent_granted=True,
            user_current_enrollment_mode='audit',
        )
```
or something like that in the test itself (or in the setup data) and then confirming the value is still what you'd expect 👍 

EDIT: I realize now that you have a note in your PR comment that tests are still incoming. Sorry for the noise~

zubair-arbi:
@brittneyexline I have made the changes now working on tests coverage. Can you please give another review on the changes?

enterprise_data/clients.py
```diff
@@ -24,6 +29,7 @@ class EnterpriseApiClient(EdxRestApiClient):
 
     API_BASE_URL = settings.LMS_BASE_URL + 'enterprise/api/v1/'
     ENTERPRISE_DATA_API_GROUP = 'enterprise_data_api_access'
+    ENTERPRISE_REPORTING_ENDPOINT = 'enterprise_customer_reporting'
```
brittneyexline: technically, we aren't hitting the enterprise reporting endpoint, so i would rename this to indicate that we are hitting with_access_to or the general enterprise-customer endpoint.

enterprise_data/filters.py
```diff
@@ -25,3 +25,26 @@ def filter_queryset(self, request, queryset, view):
...
+    variable, to identify the object's relationship to the
+    `user_current_enrollment_mode` field.
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        """
+        Filter out queryset for results where enrollment mode is `audit`.
+        """
+        enterprise_id = view.kwargs['enterprise_id']
+        enable_audit_enrollment = request.session[enterprise_id].get('enable_audit_enrollment', True)
```
brittneyexline: i wouldn't default this setting to True - it should default to false

enterprise_data/permissions.py
```diff
@@ -98,8 +98,12 @@ def has_permission(self, request, view):
             enterprise_data = self.get_enterprise_with_access_to(request.auth, request.user, enterprise_in_url)
             if not enterprise_data:
                 request.session['enterprises_with_access'][enterprise_in_url] = False
+                request.session['enable_audit_enrollment'] = True
```
brittneyexline: now that we are adding more than just authorization data in the session, we should probably extract this logic somewhere else to populate the session data with relevant enterprise data.
this assignment also looks to be inconsistent with the one below, which is specific to an enterprise, and that is inconsistent with the structure of the access data. moving this to a separate function gives us the opportunity to make this consistent.

brittneyexline (commented):
you're on the right track - i just had some suggestions for structuring the code more cleanly.

zubair-arbi:
Rebased onto master and squashed commits from has 7b09b556bbf04d191e9899ccad64b9beb2e3663e

enterprise_data/permissions.py
```diff
@@ -94,12 +95,19 @@ def has_permission(self, request, view):
 
         if ('enterprises_with_access' not in request.session or
                 enterprise_in_url not in request.session['enterprises_with_access']):
-            request.session['enterprises_with_access'] = {}
             enterprise_data = self.get_enterprise_with_access_to(request.auth, request.user, enterprise_in_url)
             if not enterprise_data:
-                request.session['enterprises_with_access'][enterprise_in_url] = False
+                update_session_with_enterprise_data(
+                    request, enterprise_in_url, enterprises_with_access=False, enable_audit_enrollment=True
```
brittneyexline: enable_audit_enrollment should default to False

brittneyexline (commented):
just one small update and you're good to go codewise! i'd also like to see a test of the filtering being applied and not applied based on what's returned from the api response.

zubair-arbi:
@saleem-latif This PR is ready for final review.

zubair-arbi:
@brittneyexline changes are finalized, please review.

saleem-latif (approved):
Looks Good :+1:

zubair-arbi:
Merging the PR now.

PR #69, opened 2018-10-12
ENT-1257

muhammad-ammar:
[ENT-1257](https://openedx.atlassian.net/browse/ENT-1257)

These changes will add support to create data reports in `json` format for only `progress_v2` data type. `progress` data type is deprecated so we don't need changes for it.

PR #71, opened 2018-10-15
Add sorting to completed courses endpoint

mushtaqak:
This adds the sorting capability to `EnterpriseLearnerCompletedCoursesViewSet`

### DIFF
--- a/enterprise_data/tests/test_views.py
+++ b/enterprise_data/tests/test_views.py
@@ -595,6 +595,7 @@ class TestEnterpriseUsersViewSet(APITestCase):
         assert response.json()['results'][3]['id'] == users[1]
 
 
+@ddt.ddt
 class TestEnterpriseLearnerCompletedCourses(APITestCase):
     """
     Tests for EnterpriseLearnerCompletedCoursesViewSet.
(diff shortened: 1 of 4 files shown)


enterprise_data/tests/test_views.py
```diff
@@ -624,3 +625,53 @@ def test_no_page_querystring_skips_pagination(self):
...
+        for ordering query param is set
+        """
+
+        # Add enrollments
+        one_day = timedelta(days=1)
+        date_in_past = timezone.now() - one_day
+        ent_user4 = EnterpriseUserFactory(
+            enterprise_user_id=4,
+        )
+        ent_user5 = EnterpriseUserFactory(
```
zubair-arbi: Instead of hard coding the number of users and their enrollments. Pass the users count `2` as test data and create users/enrollments according to that.

enterprise_data/tests/test_views.py
```diff
@@ -624,3 +625,53 @@ def test_no_page_querystring_skips_pagination(self):
...
+            has_passed=True,
+            consent_granted=True,
+        )
+        url = reverse(
+            'v0:enterprise-learner-completed-courses-list',
+            kwargs={'enterprise_id': self.enterprise_id}
+        )
+        params = {'ordering': ordering}
+        response = self.client.get(url, params)
+        assert response.json()['count'] == 2
```
zubair-arbi: Assert this with users count `2` from test data as suggested above.

zubair-arbi (changes requested):

enterprise_data/tests/test_views.py
```diff
@@ -624,3 +625,112 @@ def test_no_page_querystring_skips_pagination(self):
...
+                    consent_granted=True,
+                )
+        url = reverse(
+            'v0:enterprise-learner-completed-courses-list',
+            kwargs={'enterprise_id': self.enterprise_id}
+        )
+        params = {'ordering': ordering}
+        response = self.client.get(url, params)
+        assert response.json()['count'] == expected_results_count
+        import pdb; pdb.set_trace();
```
irfanuddinahmad: Please remove

irfanuddinahmad (approved):
LGTM

enterprise_reporting/__init__.py
```diff
@@ -4,4 +4,4 @@
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.2.5"
+__version__ = "0.2.15"
```
mushtaqak: @muhammad-ammar We need version bump on `enterprise_reporting/__init__.py` for your https://github.com/edx/edx-enterprise-data/pull/69 PR, right? cc: @zubair-arbi

mushtaqak: @georgebabey Do you know we need version bump in  `enterprise_reporting/__init__.py` ? and  reason behind versioning in 2 paths.

georgebabey: Ugh. yes. bit of history - before there was only the enterprise_reporting directory with python scripts. Then when we created the enterprise_data django app I decided this was the best repo thinking we'd move the reporting scripts over as management commands. We never the reporting scripts so now we have 2 different directories in the same repo.

Because enterprise_reporting doesn't get published it technically doesn't need to be updated. The published version comes from enterprise_data folder: https://github.com/edx/edx-enterprise-data/blob/master/setup.py#L25

mushtaqak: Great. Since enterprise_reporting doesn't get published, updating version in this directory won't have any effect, so I will just remove this.

enterprise_data/tests/test_views.py
```diff
@@ -652,3 +653,111 @@ def test_no_page_querystring_skips_pagination(self):
...
+                    has_passed=True,
+                    consent_granted=True,
+                )
+        url = reverse(
+            'v0:enterprise-learner-completed-courses-list',
+            kwargs={'enterprise_id': self.enterprise_id}
+        )
+        params = {'ordering': ordering}
+        response = self.client.get(url, params)
+        assert response.json()['count'] == expected_results_count
```
zubair-arbi, line 760: move these asserts in loop too.

zubair-arbi (approved):
Just one small suggestion, otherwise LGTM 👍

PR #75, opened 2018-10-19
ENT-1304 - Adding sorting capabilities to extra_fields values

christopappas:
This PR allows the `enrollment_count` and `course_completion_count` to be sortable by doing the following:

- moves the calculation of `enrollment_count` and `course_completion_count` from the serializer to the view, because sorting must be done at the layer at which you are "grabbing all the stuff you need" from the database (aka the view's `get_queryset` method)
- annotates the EnterpriseUser queryset with the above count values if query params are passed in the request
- makes the serializer check for the appropriate query params so it knows to expose the calculated value in the object representation

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,9 @@ Change Log
 
 Unreleased
 ----------
+[1.0.1] - 2018-10-23
+--------------------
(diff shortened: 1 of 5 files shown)


adamstankiewicz (approved):
Nice job adding tests for sorting, covering both asc/desc sort directions.

georgebabey (approved):

PR #76, opened 2018-10-24
asadiqbal08/ENT-1232 Unenrollments in Learner Reports 

asadiqbal08:
Unenrollments in Learner Reports [ENT-1232](https://openedx.atlassian.net/browse/ENT-1232)

As an Enterprise customer, I want to be able to have my employees unenroll in a course within the 14 day refund period so they have the opportunity to see if the course is or is not the right one for their needs.
As an edX operator, when a learner unenrolls, I want to know the enrollment date, and whether it occurred within the refund period


enterprise_data/api/v0/serializers.py
```diff
@@ -14,13 +14,30 @@ class EnterpriseEnrollmentSerializer(serializers.ModelSerializer):
...
+        """
+        if obj.unenrollment_timestamp:
+            period = range(0, 15)
+            unenrollment_within_date = (obj.unenrollment_timestamp - obj.course_start).days in period or \
+                                       (obj.unenrollment_timestamp - obj.enrollment_created_timestamp).days in period
+
+            if unenrollment_within_date:
+                return True
+
+        return False
```
brittneyexline: i think it would be better to return None when there is no unenrollment to avoid confusion.

enterprise_data/api/v0/serializers.py
```diff
@@ -14,13 +14,30 @@ class EnterpriseEnrollmentSerializer(serializers.ModelSerializer):
...
         )
 
+    def get_unenrollment_end_within_date(self, obj):
+        """
+        Return "True" when un-enrolled date is within 14 days of course start or enrollment date
+        (whichever is later), when un-enrolled date is after 14 days of
+        course start or enrollment date (whichever is later) return "False".
+        """
+        if obj.unenrollment_timestamp:
+            period = range(0, 15)
```
brittneyexline: it might be good to define this as a constant, in case we change our policy.

enterprise_data/api/v0/serializers.py
```diff
@@ -14,13 +14,30 @@ class EnterpriseEnrollmentSerializer(serializers.ModelSerializer):
...
+            unenrollment_within_date = (obj.unenrollment_timestamp - obj.course_start).days in period or \
```
brittneyexline: nit: we could just use simple comparisons (>0, <=14) instead of checking if an integer is in a list. it would be slightly more performant and readable imo.

brittneyexline (commented):
just a few small things to address, plus adding tests!

asadiqbal08:
@brittneyexline kindly take another look.

enterprise_data/api/v0/serializers.py
```diff
@@ -14,13 +14,29 @@ class EnterpriseEnrollmentSerializer(serializers.ModelSerializer):
...
         )
 
+    def get_unenrollment_end_within_date(self, obj):
+        """
+        Return "True" when un-enrolled date is within 14 days of course start or enrollment date
+        (whichever is later), when un-enrolled date is after 14 days of
+        course start or enrollment date (whichever is later) return "None".
+        """
+        if obj.unenrollment_timestamp:
+            unenrollment_within_date = (0 < (obj.unenrollment_timestamp - obj.course_start).days <= 14) or \
```
brittneyexline, line 32: shouldn't we just return whatever this evaluates to? as is, we would return None when we probably should return False

brittneyexline (commented):
i think adding test cases will be good to clarify the logic you have in the serializer. also keep in mind that this should not be merged until you analytics pipeline PR has merged and run in production.

enterprise_data/api/v0/serializers.py
asadiqbal08: I prefer to return either `True` or `False` instead of `None`. I agree, we just return whatever this evaluates to `True/False`

enterprise_reporting/export_edxapp_data.py
```diff
@@ -34,9 +34,10 @@
 import sys
 from io import open  # pylint: disable=redefined-builtin
 
-import mysql.connector
 import vertica_python
 
+import mysql.connector
```
zubair-arbi: Can you please revert this files change? as it seems irrelevant.

zubair-arbi (approved):
Just one small suggestion otherwise LGTM 👍

enterprise_reporting/export_edxapp_data.py
asadiqbal08: The change here is showing by `isort` command. That import is in use.

asadiqbal08:
@brittneyexline your final review in it ...

enterprise_data/api/v0/serializers.py
```diff
@@ -14,13 +14,27 @@ class EnterpriseEnrollmentSerializer(serializers.ModelSerializer):
...
             enterprise_id=obj.enterprise_id, course_id=obj.course_id
         )
 
+    def get_unenrollment_end_within_date(self, obj):
+        """
+        Return "True" when un-enrolled date is within 14 days of course start or enrollment date
+        (whichever is later), when un-enrolled date is after 14 days of
+        course start or enrollment date (whichever is later) return "False".
+        """
+        unenrollment_within_date = False
```
brittneyexline: i think this should default to None if there is no unenrollment timestamp. i know you disagreed earlier, but to me, it is misleading for it to be false when there is no unenrollment timestamp to evaluate. why not make each case return a distinct value instead of overloading false?

enterprise_data/fixtures/enterprise_enrollment.json
```diff
@@ -34,7 +34,8 @@
       "current_grade": 0.80,
       "course_price": 200.00,
       "discount_price": 120.00,
-      "created": "2018-08-01 23:14:35Z"
+      "created": "2018-08-01 23:14:35Z",
+      "unenrollment_timestamp": "2014-06-29 16:02:38Z"
```
brittneyexline: setting unenrollment timestamp for all the fixture data means we are missing the case where the user has not unenrolled, which is likely to be the most common case. you should leave at least one of these rows with unenrollment timestamp being null.

asadiqbal08:
@brittneyexline Updated

brittneyexline (approved):

brianhw:
Is this fixture supposed to exactly match the table that is generated by the pipeline in the resultstore?   This PR sets up the table to have the unenrollment_timestamp field appear after the created field, and that's not the case.   The 'created' field is always last.

asadiqbal08:
@brianhw this [fixture](https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/fixtures/enterprise_enrollment.json) is used in development process for testing. So what do you suggest, should I create a new PR with re-order the columns structure as you mentioned the "created" field is always be the last in order  or should do in separate ticket ?

PR #77, opened 2018-10-24
ENT-1318 | Updating course_completion_count subquery logic so it does not return

christopappas:
Updating course_completion_count subquery logic so it does not return "1" when more than 1 course has been completed. also updated enrollmentFactory so the ids clash less for enrollment test fixtures


georgebabey (approved):
Makes sense, good catch

PR #80, opened 2018-10-25
ENT-1325 | Updating enrollment_count and course_completion_count logic

christopappas:
ENT-1325 | Updating enrollment_count and course_completion_count computations to restrict to consent_granted=True enrollments

What I've done here:

- In EnterpriseUsersViewSet, I've changed the original filter of users to no longer include enrollments with `consent_granted=None`.
- I've also dropped the references to `CONSENT_TRUE_OR_NONE_Q` in other places within the viewset, and replaced it with a simpler Q object. Using `CONSENT_TRUE_OR_NONE_Q` everywhere added potentially redundant predicates to the SQL the ORM would create, and given the already complex nature of what we're filtering, I wanted to make sure we're doing as little filtering as possible to get what we need.
- Related to the actual ticket: `enrollment_count` is now calculated using a subquery just as `course_completion_count` is
- `course_completion_count` subquery now filters on `consent_granted=True` instead of excluding `consent_granted=False`
- I've added another test for `learner_activity_filter` as a proactive measure to verify other areas of our code are doing what we expect with respect to `consent_granted`

### DIFF
--- a/enterprise_data/tests/test_views.py
+++ b/enterprise_data/tests/test_views.py
@@ -205,6 +205,52 @@ class TestEnterpriseEnrollmentsViewSet(APITestCase):
         for enrollment in result['results']:
             assert datetime.strptime(enrollment['last_activity_date'], "%Y-%m-%d").date() in expected_dates
 
+    @ddt.data(
+        (
+            'active_past_week',
+            [date.today(), date.today() - timedelta(days=2)]
+        ),
+        (
+            'inactive_past_week',
+            [date.today() - timedelta(weeks=2), subtract_one_month(date.today())]
+        ),
+        (
+            'inactive_past_month',
+            [subtract_one_month(date.today())]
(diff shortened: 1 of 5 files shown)


enterprise_data/tests/test_views.py
```diff
@@ -298,6 +298,12 @@ def setUp(self):
             enterprise_user=self.ent_user4,
             course_end=date_in_future,
         )
+        EnterpriseEnrollmentFactory(
+            enterprise_user=self.ent_user4,
+            course_end=date_in_past,
+            consent_granted=True,
```
christopappas, line 355: I decided to add an enrollment for this user in the test data to correct the test asserts (after making my changes) instead of changing the assertion to expect 1 (because in the past expecting 1 did not uncover a bug)

georgebabey (approved):

PR #81, opened 2018-10-29
Fixed PGP encryption key issue

irfanuddinahmad:
This PR sets the PGP encryption key to None if the key is not found in the reporting configuration.


zubair-arbi (approved):
LGTM 👍

mushtaqak (approved):
Let's ship

PR #84, opened 2018-11-01
ENT-1336. Handle case when pgp_key is empty string

georgebabey:
The pgp_key config in the LMS is a TextField which means when it is updated the key is set to empty string ('') and not None. We were only checking for None and therefore passing along an empty pgp_key to the _get_encrypted_file method. This is fixed with this PR by doing a truthy check - of which both empty string and None will fail.

I also updated the default to better match what is getting passed in, as well as make the tests fail before fixing the problem.

### DIFF
--- a/enterprise_reporting/utils.py
+++ b/enterprise_reporting/utils.py
@@ -34,13 +34,13 @@ FREQUENCY_TYPE_WEEKLY = 'weekly'
 AWS_REGION = 'us-east-1'
 
 
-def compress_and_encrypt(files, password=None, pgp_key=None):
+def compress_and_encrypt(files, password=None, pgp_key=''):
     """
     Given file(s) and a password or a PGP key,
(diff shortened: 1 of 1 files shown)


georgebabey:
Based on the guidance on the django docs, we should avoid nulls entirely for string-based fields, however this would require an update to the edx-enterprise package and a migration on the LMS. I think we should still consider that in order to keep the database tidy, but this will fix the immediate issue by handling both null/None and empty string.

> Avoid using null on string-based fields such as CharField and TextField. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL

https://docs.djangoproject.com/en/2.1/ref/models/fields/#django.db.models.Field.null

irfanuddinahmad (approved):
LGTM

PR #89, opened 2018-11-06
Update pgp files to include base extension

georgebabey:
Rather than replace the current extension, this tacks on .pgp to the end. This will give a better indication of the file type to our customers after it has been decrypted.


georgebabey:
Tested via jenkins - https://tools-edx-jenkins.edx.org/job/enterprise/job/enterprise-data-reporting/9250

brittneyexline (approved):

PR #90, opened 2018-11-27
ENT-1244 | External Resource Link Report

christopappas:
Report generation script that calls coursegraph for data, creates csv, and emails it.

Please ask any questions that come to mind!

Notes:
- `py2neo` is very convenient, but added a few [most likely unimpactful, yet never the less more] dependencies.
- The first pass at filtering links comes down to knocking out images and obvious edx extensions.
- ~Needs tests? Yep. I think I have another PR that adds the `enterprise_reporting` app to the tests which will be needed if we want this code to have tests that run with `make validate`.~ Never got to this. I would like to add in test coverage for the reporting app as a whole within the next month or so

### DIFF
--- a/enterprise_reporting/external_resource_link_report.py
+++ b/enterprise_reporting/external_resource_link_report.py
@@ -1,0 +1,154 @@
+# -*- coding: utf-8 -*-
+"""
+External Resource Link Report Generation Code.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from datetime import date
+import logging
+import os
+import re
+import sys
+
+from py2neo import Graph
+
+from enterprise_reporting.utils import send_email_with_attachment
+
+logging.basicConfig(level=logging.INFO)
+LOGGER = logging.getLogger(__name__)
+
+
+def generate_csv_string(processed_results):
+    """
+    Takes a dict of processed results and turns it into a string suitable
+    to be written as a csv file
+
+    Returns (unicode) string
+    """
+    csv_string = u'Course Key,Course Title,Partner,External Links\n'
+    for course_key, data in processed_results.items():
+        links = u'\n,,,'.join(data['external_links'])
+        csv_string += u'{},"{}",{},{}\n'.format(
+            course_key,
+            data['course_title'],
+            data['organization'],
+            links,
+        )
+    return csv_string
+
+
+def gather_links_from_html(html_string):
+    """
+    Takes some html blob as a string and extracts any external links, and
+    returns them as a set
+    """
+    pattern = 'https?://.*?[" <]'
+    links = set([
+        link[0:-1]
+        for link in re.findall(pattern, html_string,)
+        if (not link.lower().endswith('.png"') and
+            not link.lower().endswith('.jpg"') and
+            not link.lower().endswith('.jpeg"') and
+            not link.lower().endswith('.gif"') and
+            not '.edx.org' in link)
+    ])
+    return links
+
+
+def process_results(raw_results):
+    """
+    Takes the data from a coursegraph query
+
+    Returns a dict with course keys as the key and list data about that course
+    as value
+    """
+    processed_results = {}
+    for entry in raw_results:
+        course_key = entry['h.course_key']
+        # Only want new style course keys to exclude archived courses
+        if not course_key.startswith('course-'):
+            continue
+
+        external_links = gather_links_from_html(entry['h.data'])
+
+        if not external_links:
+            continue
+
+        if course_key not in processed_results:
+            processed_results[course_key] = {
+                'course_title': entry['course_title'],
+                'organization': entry['organization'],
+                'external_links': external_links,
+            }
+        else:
+            processed_results[course_key]['external_links'].update(external_links)
+    return processed_results
+
+
+def query_coursegraph():
+    """
+    Calls coursegraph with cypher query and returns query data
+
+    The data is a list of dicts where each dict contains keys that correspond
+    to what is being returned in the query
+    """
+    graph = Graph(
+        bolt=True,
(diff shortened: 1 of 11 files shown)


brittneyexline (commented):
looks good overall! i would say next steps would be adding some logging and error handling, and moving some of the strings you have to constants. i also remember you saying that coursegraph only gets updated once per week, so if someone runs this on demand, should we consider having it send a file that was written to disk by a previous run if available to avoid hitting coursegraph again?

enterprise_reporting/external_resource_link_report.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def generate_csv_string(processed_results):
+    """
+    Takes a dict of processed results and turns it into a string suitable
+    to be written as a csv file
+
+    Returns (unicode) string
+    """
+    csv_string = u'Course Key\tExternal Links\n'
```
georgebabey: Should we be using commas here instead of a tab-delimited csv? And if so we'll need to escape the commas in the links

enterprise_reporting/external_resource_link_report.py
```diff
@@ -0,0 +1,138 @@
...
+def gather_links_from_html(html_string):
+    """
+    Takes some hmtl blob as a string and extracts any external links, and
+    returns them as a set
+    """
+    pattern = 'https?://.*?[" <]'
+    links = set([
+        link[0:-1]
+        for link in re.findall(pattern, html_string,)
+        if (not link.lower().endswith('.pdf"') and
```
georgebabey: I think we should not exclude .pdf, as there may be external pdfs needed for the course.

enterprise_reporting/external_resource_link_report.py
```diff
@@ -0,0 +1,138 @@
...
+            not link.lower().endswith('.png"') and
+            not link.lower().endswith('.jpg"') and
+            not link.lower().endswith('.jpeg"') and
+            not link.lower().endswith('.gif"') and
+            not '.edx.org' in link and
+            not '.edx.com' in link)
```
georgebabey: Are there really edx.com links? 😕

enterprise_reporting/external_resource_link_report.py
```diff
@@ -0,0 +1,138 @@
...
+    raw_results = query_coursegraph()
+
+    # Do additional processing/clean up results
+    processed_results = process_results(raw_results)
+
+    # Generate csv string
+    csv_string = generate_csv_string(processed_results)
+
+    subject = 'External Resource Link Report'
+    body = '''Dear Customer Support,
```
georgebabey: s/Support/Success/

enterprise_reporting/external_resource_link_report.py
```diff
@@ -0,0 +1,138 @@
...
+Find attached a file containing course keys and their respective
+external resource links.
+
+If you have any questions/concerns with the report, please ask the
+Enterprise Team (kindly)!
+
+Sincerely,
+The Enterprise Team'''
+    from_email = 'someEmail@edx.org'
+    to_email = 'someEmail@edx.org' #change me to param
```
georgebabey: Don't forget these todos!

enterprise_reporting/external_resource_link_report.py
```diff
@@ -0,0 +1,138 @@
...
+        secure=True,
+    )
+    query = '''MATCH
+                (c:course)-[:PARENT_OF*]->(h:html) 
+              WHERE 
+                h.data =~ '.*https?://.*'
+              RETURN 
+                h.course_key, 
+                h.data, 
+                "https://studio.edx.org/container/" + h.location as link
```
georgebabey: Looks like you're only using `course_key` and `data` so you could remove `link` (which is also confusing since links above refers to external links, but this is a link to the course.

georgebabey (changes requested):
Looks good, just a few minor suggestions

enterprise_reporting/external_resource_link_report.py
christopappas: I admit, tabs are my delimiter personal preference for creating spreadsheets, and I can confirm it opens in excel alright (also prints nicely!). I'm happy to change if you think it should be though

enterprise_reporting/external_resource_link_report.py
christopappas: Okay, can remove this

enterprise_reporting/external_resource_link_report.py
christopappas: Did another check. apparently not. i could have sworn I saw something earlier on in my investigation. I'll remove this as it is probably unnecessary

enterprise_reporting/external_resource_link_report.py
christopappas: well sed. will change

enterprise_reporting/external_resource_link_report.py
christopappas: will do

brianhw:
If you're interested in generating output suitable for Excel, you might want to consider using the "csv" package included in Python.   In particular, I understand they have a dialect='excel' option for the csv output that would handle decisions about delimiters and escaping and all that.

christopappas:
@brianhw That is a good recommendation. I ended up going with a custom string handling solution as the consumers of this report wanted some special whitespace formatting. It is a csv insofar that the actual column delimiters are commas (with quotes being used to escape any commas in titles), but the desire was for this report potentially to be showed to customers as-is, and therefore, was desired to look as though it was "pretty printed" to some extent

If I need to make further modifications to make this more like a standard CSV, I will certainly give the csv package a whirl 👍

georgebabey (approved):

PR #91, opened 2018-12-06
ENT-1026 | Updating report to use catalog uuid that comes from lms api call

christopappas:
Also adds basic test filesfor the eneterprise_reporting app

NOTE: All uuids in fixture file re from my local setup, so nothing secret there


georgebabey (approved):
This is great and thank you for getting the testing framework in 🥇

PR #100, opened 2019-01-02
ENT-1396 | Updating External Links Report to also include an aggregate report or domains

christopappas:
This change not only updates the way the external link report works, but also update the send_email_with_attachments function to allow lists of string data to be used as multiple attachments.

### DIFF
--- a/enterprise_reporting/external_resource_link_report.py
+++ b/enterprise_reporting/external_resource_link_report.py
@@ -4,11 +4,17 @@ External Resource Link Report Generation Code.
 """
 from __future__ import absolute_import, unicode_literals
 
+from collections import Counter
+from datetime import date
+import operator
 import logging
 import os
 import re
 import sys
-from datetime import date
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
 
 from py2neo import Graph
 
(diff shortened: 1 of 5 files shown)


enterprise_reporting/external_resource_link_report.py
```diff
@@ -55,11 +89,15 @@ def gather_links_from_html(html_string):
...
 
-def process_results(raw_results):
+def process_coursegraph_results(raw_results, domains_and_counts=False):
     """
     Takes the data from a coursegraph query
 
-    Returns a dict with course keys as the key and list data about that course
+    domains_and_counts - specfies that urls should be stripped down to only
+                         their domain, and then a count of occurences is also
+                         added to processed_results dixt
```
brittneyexline: nit: dixt -> dict

enterprise_reporting/external_resource_link_report.py
```diff
@@ -74,14 +112,37 @@ def process_results(raw_results):
         if not external_links:
             continue
 
-        if course_key not in processed_results:
-            processed_results[course_key] = {
-                'course_title': entry['course_title'],
-                'organization': entry['organization'],
-                'external_links': external_links,
-            }
+        if domains_and_counts:
```
brittneyexline: why not remove the domains_and_counts and do all of the calculations in one go? you could rename external links here to domain_counts and keep external links for the original use.

enterprise_reporting/external_resource_link_report.py
```diff
@@ -18,7 +24,35 @@
...
-def generate_csv_string(processed_results):
+def generate_aggregate_report_csv_string(processed_results):
+    """
+    Takes a dict of processed results and turns it into a string suitable
+    to be written as a csv file
+
+    Returns (unicode) string
+    """
+    csv_string = u'Course Key,Course Title,Partner,External Domain,Count\n'
+    for course_key, data in processed_results.items():
```
brittneyexline, line 45: i wonder if there is a way to combine some of this logic with the function below, so maybe these both call to a common function that handles the common logic?

brittneyexline (commented):
looks good generally, though i think there are some opportunities to consolidate some of the code.

enterprise_reporting/external_resource_link_report.py
christopappas: Hmm... could I bother you to elaborate a little more on this? Not quite sure what you have in mind

If this did all the calculations in one go, what would this function return?

(Semi related, `external_links = gather_links_from_html(entry['h.data'])` returns a set... So if the same url shows up in the same html chunk more than once, it would be counted only as one. I'm not sure how likely that is to happen, but I should fix that)

enterprise_reporting/external_resource_link_report.py
christopappas: I originally decided to keep things separate because conceptually it made sense to me, as there are 2 separate reports. I can see what I can do though

enterprise_reporting/external_resource_link_report.py
```diff
@@ -43,23 +81,27 @@ def gather_links_from_html(html_string):
...
             not link.lower().endswith('.jpeg"') and
             not link.lower().endswith('.gif"') and
             not '.edx.org' in link)
-    ])
+    ]
     return links
 
 
-def process_results(raw_results):
+def process_coursegraph_results(raw_results, domains_and_counts=False):
```
brittneyexline: i think you can get rid of domains_and_counts now

enterprise_reporting/utils.py
```diff
@@ -72,32 +72,71 @@ def _get_compressed_file(files, password=None):
...
 
     Adapted from https://gist.github.com/yosemitebandit/2883593
 
-    attachment_data should be string data if value is passed in.
-    If filename refers to a file that lives on the filesystem instead
-    of just the name you want the attachment to have, no need to
-    set attachment_data to anything. This gives us the option of
-    handing this function string data that lives in memory instead
-    of needing to first write to a file.
+    filename can be a string or list of strings
```
brittneyexline: i'd rather see filename expect one data type (list), and maybe attachment_data should be the same. it would be good to have some consistency with what data types get passed in so there is not so much logic to "resolve" attachments. this will probably mean adjusting callers, and i don't know how difficult that would be. if it turns into a rabbit hole, don't bother.

brittneyexline (commented):
closer! the csv generation looks better - i think we can simplify the attachment logic though.

enterprise_reporting/utils.py
christopappas: imagine this: 

```
file_dict = {
     'my_file_name.csv': '1,1,1,1,1,\n2,2,2,2,2\n3,3,3,3,3\n'
     '/abs/path/to/myfile.csv': None
}
```
Instead of handing both filename and attachment_data, I hand something like this. If the value is `None`, we look on the filesystem, else we use the string data in the value as the attachment

what do you think? :)

brittneyexline (approved):

PR #101, opened 2019-01-03
ENT-1440. Fixed serializer exception when calculating unenroll and course start…

georgebabey:
… is None

### DIFF
--- a/enterprise_data/api/v0/serializers.py
+++ b/enterprise_data/api/v0/serializers.py
@@ -29,8 +29,11 @@ class EnterpriseEnrollmentSerializer(serializers.ModelSerializer):
         """
         unenrollment_within_date = None
         if obj.unenrollment_timestamp:
-            unenrollment_within_date = (0 < (obj.unenrollment_timestamp - obj.course_start).days <= 14) or \
-                                       (0 < (obj.unenrollment_timestamp - obj.enrollment_created_timestamp).days <= 14)
+            unenrollment_within_date = (
+                not obj.course_start or
+                0 < (obj.unenrollment_timestamp - obj.course_start).days <= 14 or
+                0 < (obj.unenrollment_timestamp - obj.enrollment_created_timestamp).days <= 14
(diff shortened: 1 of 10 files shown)


georgebabey:
Also ran make upgrade for some minor version bumps that I just kept in

brittneyexline (approved):

enterprise_data/api/v0/serializers.py
```diff
@@ -29,7 +29,8 @@ def get_unenrollment_end_within_date(self, obj):
         """
         unenrollment_within_date = None
         if obj.unenrollment_timestamp:
-            unenrollment_within_date = (0 < (obj.unenrollment_timestamp - obj.course_start).days <= 14) or \
+            unenrollment_within_date = (not obj.course_start) or \
```
christopappas: Some style nitpicks, but, parentheses are not required to be around the `not obj.course_start`, although I can see parentheses making the longer expressions clearer (though I'm fairly certain they too are not necessary).

Also, I personally don't love backslashes. I usually evangelize use of parens as in, something like (untested):

```
unenrollment_within_date = (not obj.course_start or 
                                                (0 < (obj.unenrollment_timestamp - obj.course_start).days <= 14) or
                                                (0 < (obj.unenrollment_timestamp - obj.enrollment_created_timestamp).days <= 14))
```
or if those lines are too long:
```
unenrollment_within_date = (
    not obj.course_start or 
    0 < (obj.unenrollment_timestamp - obj.course_start).days <= 14 or
    0 < (obj.unenrollment_timestamp - obj.enrollment_created_timestamp).days <= 14
)
```

Just some suggestions anyway

christopappas (approved):
Looks good, just made a style suggestion in case you like that better

enterprise_data/api/v0/serializers.py
georgebabey: Was just following the existing style, but agree and will update

PR #104, opened 2019-01-09
ENT-1443 | Adding management commands for dev data convenience

christopappas:
2 new management commands are being added. One for creating EnterpriseUsers, the other for EnterpriseEnrollments

This also incorporates a number of pylint cleanups

Will bump the version when it gets approved!

### DIFF
--- a/enterprise_data/management/commands/create_enterprise_enrollment.py
+++ b/enterprise_data/management/commands/create_enterprise_enrollment.py
@@ -1,0 +1,51 @@
+"""
+management command for creating enterprise enrollments
+"""
+from __future__ import absolute_import, unicode_literals
+
+import sys
+
+from django.core.management.base import BaseCommand, CommandError
+
+import enterprise_data.tests.test_utils
+
+
+class Command(BaseCommand):
+    """ management command class """
+    help = 'Creates an EnterpriseEnrollment with randomized attributes'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'enterprise_id',
+            type=str,
+            help='UUID for an enterprise'
+        )
+        parser.add_argument(
+            'enterprise_user_id',
+            type=int,
+            default=None,
+            help='enterprise_user_id for an enterprise_user'
+        )
+
+    def handle(self, *args, **options):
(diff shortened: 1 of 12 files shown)


enterprise_data/management/commands/create_enterprise_enrollment.py
```diff
@@ -0,0 +1,51 @@
...
+                enterprise_user_id=enterprise_user_id,
+            )
+            info = (
+                '\nCreated EnterpriseEnrollment with enterprise_id '
+                '{} for EnterpriseUser with enterprise_user_id of {}\n\n'.format(
+                    enterprise_id,
+                    enterprise_user_id,
+                )
+            )
+            sys.stdout.write(info)
```
brittneyexline, line 45: any reason we're not using a logger?

enterprise_data/management/commands/create_enterprise_enrollment.py
```diff
@@ -0,0 +1,51 @@
...
+
+import enterprise_data.tests.test_utils
+
+
+class Command(BaseCommand):
+    """ management command class """
+    help = 'Creates an EnterpriseEnrollment with randomized attributes'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
```
brittneyexline, line 18: any reason not to have names for these options instead of relying on position? i think if we expand these commands to allow writing a specific value for the field, we would want to specify options by a name

brittneyexline (commented):
just a few small questions regarding the commands!

enterprise_data/management/commands/create_enterprise_enrollment.py
christopappas: I originally had some trouble actually setting up the logger, but then realized that this was really just a productivity tool for devs locally, so having a proper logger was not necessary.

enterprise_data/management/commands/create_enterprise_enrollment.py
christopappas: I think this is a good thought, but I personally would like to punt on that until we decide to expand the functionality of these commands

brittneyexline (approved):

PR #106, opened 2019-01-23
ENT-1455: return empty list if no enrollments

bessiesteinberg:
- the ```/enterprise/api/v0/enterprise/<enterprise_uuid>/enrollments```
  used to return 404 if there were no enrollments for that enterprise
- now that case will return an empty list to signify that there is an
  enterprise with that UUID, but it has no enrollments
- if there is no enterprise user with the given UUID, 404 is thrown and
  the following error message is logged:
  ```No enterprise found with id {enterprise_id} from endpoint '{path}'.```

Remove 404 and return empty list when enterprise has 0 enrollments. Log error and throw 404 when enterprise doesn't exist


georgebabey (approved):
Looks good - thanks for cleaning up the tests and typo!

PR #109, opened 2019-02-28
ENT-1603: Only filter out audit rows that do not have any offer or code applied

saleem-latif:
**JIRA Ticket:** [ENT-1603](https://openedx.atlassian.net/browse/ENT-1603)

**Description:**
When users unenroll, even if their original enrollment mode was verified, the enrollment mode gets set to audit. When we are reporting from the data API, we are filtering out audit rows depending on if the enterprise is configured to include or exclude audit enrollments (https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/filters.py#L41). As a short term fix, we should update this logic to only filter out audit rows that do not have any offer or code applied.

**Acceptance Criteria:**

- Only filters out audit rows that do not have any offer or code applied

### DIFF
--- a/enterprise_data/filters.py
+++ b/enterprise_data/filters.py
@@ -55,7 +55,12 @@ class AuditEnrollmentsFilterBackend(filters.BaseFilterBackend):
         enable_audit_enrollment = request.session['enable_audit_enrollment'].get(enterprise_id, False)
 
         if not enable_audit_enrollment:
-            filter_kwargs = {view.ENROLLMENT_MODE_FILTER: 'audit'}
-            queryset = queryset.exclude(**filter_kwargs)
+            # Filter out enrollments that have audit mode and do not have a coupon code or an offer.
+            filter_query = {
+                view.ENROLLMENT_MODE_FILTER: 'audit',
+                view.COUPON_CODE_FILTER: None,
+                view.OFFER_FILTER: None
+            }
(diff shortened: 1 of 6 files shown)


enterprise_data/filters.py
```diff
@@ -55,7 +55,11 @@ def filter_queryset(self, request, queryset, view):
         enable_audit_enrollment = request.session['enable_audit_enrollment'].get(enterprise_id, False)
 
         if not enable_audit_enrollment:
-            filter_kwargs = {view.ENROLLMENT_MODE_FILTER: 'audit'}
-            queryset = queryset.exclude(**filter_kwargs)
+            filter_query = Q(
```
brittneyexline: could you add a comment briefly explaining this filter query and why we're doing it?

brittneyexline (commented):
looks good so far, would be good to add unit test covering the different cases affected by adding this query.

saleem-latif:
@brittneyexline I have incorporated your feedback, Also I had to update the filter criteria (Since, we are excluding records `OR` should be replaced by `AND`) after adding test cases to fix an issue. Can you take another look please?

brittneyexline (approved):

muhammad-ammar (approved):
LGTM 👍

irfanuddinahmad (approved):
LGTM!

PR #110, opened 2019-03-15
Feature role models for rule/role based permissions

muhammad-ammar:


enterprise_data/__init__.py
```diff
@@ -7,3 +7,9 @@
 __version__ = "1.0.17"
 
 default_app_config = "enterprise_data.apps.EnterpriseDataAppConfig"  # pylint: disable=invalid-name
+
+
+def update_settings(django_settings):
+    """Set enterprise_data settings."""
+    django_settings.INSTALLED_APPS = django_settings.INSTALLED_APPS + ("rules.apps.AutodiscoverRulesConfig",)
+    django_settings.AUTHENTICATION_BACKENDS.insert(0, "rules.permissions.ObjectPermissionBackend")
```
muhammad-ammar: **Note to Reviewers:** Changes in this file and `enterprise_data/apps.py` are required so that when `edx-enterprise-data` pkg is installed in `edx-analytics-data-api`, `INSTALLED_APPS` and `AUTHENTICATION_BACKENDS` are automatically updated when edx-analytics-data-api` server runs. 

This approach is borrowed from edx-platform, please see https://github.com/edx/edx-platform/blob/53458cd3eac14bc3d70257ab2d91418a82c3d807/common/djangoapps/third_party_auth/apps.py#L22 and https://github.com/edx/edx-platform/blob/53458cd3eac14bc3d70257ab2d91418a82c3d807/common/djangoapps/third_party_auth/settings.py#L16-L96

enterprise_data/rules.py
```diff
@@ -0,0 +1,24 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@rules.predicate
+def has_data_api_django_group_access(user, obj):
+    """
+    Returns whether the user belongs to `enterprise_data_api_access` group.
+    """
+    return True
```
muhammad-ammar: @brittneyexline  [EnterpriseLearnerCompletedCoursesViewSet](https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/api/v0/views.py#L330) and other views uses [HasDataAPIDjangoGroupAccess](https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/permissions.py#L70). In that permission class, We are making a request to LMS to check if the request has access to `enterprise_data_api_access` group and also update the request session. My question is are we going to add/move all the logic `HasDataAPIDjangoGroupAccess` into the `predicate`?

muhammad-ammar:
@brittneyexline  For the [EnterpriseLearnerCompletedCoursesViewSet](https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/api/v0/views.py#L330)  viewset, I don't think `permission_required` decorator will work? What about adding the complete https://github.com/escodebar/django-rest-framework-rules pkg in edx_rbac… it has `PermissionRequiredMixin` which I think can be used for all the endpoint in a viewset? Any thoughts on how we should handle this?

```python
class EnterpriseLearnerCompletedCoursesViewSet(EnterpriseViewSet, viewsets.ModelViewSet):
    """
    View to manage enterprise learner completed course enrollments.
    """
    serializer_class = serializers.LearnerCompletedCoursesSerializer
    filter_backends = (filters.OrderingFilter,)
    ordering_fields = '__all__'
    ordering = ('user_email',)

    def get_queryset(self):
        """
        Returns number of completed courses against each learner.
        """
        enterprise_id = self.kwargs['enterprise_id']
        # Get the number of completed courses against a learner.
        enrollments = EnterpriseEnrollment.objects.filter(
            enterprise_id=enterprise_id,
            has_passed=True,
            consent_granted=True,
        ).values('user_email').annotate(completed_courses=Count('course_id')).order_by('user_email')
        return enrollments
```

muhammad-ammar:
@brittneyexline This is WIP PR, I have some questions added in comments. Can you please take a look?

muhammad-ammar:
@brittneyexline One more question, We have [JwtAuthCookieRoleMiddleware](https://github.com/edx/ecommerce/pull/2148/files?file-filters%5B%5D=.py#diff-d813067575c6d20090ebe7dd78d48920) in [authz demo](https://github.com/edx/ecommerce/pull/2148). Should we need to implement/use this middleware for `edx-enterprise-data` or this be a part of `edx_rbac`?

enterprise_data/__init__.py
brittneyexline: after some thought, i think it would be better to update the analytics api directly to add these settings instead of having this package add them. this is because we can reasonably expect other features within the service to use django rules. although we are the only ones using it today, we expect more use cases to utilize this approach.

you'll have to bump the version in the analytics api, so i would remove this settings logic here and add it to the analytics api code directly when you do the version bump PR.

brittneyexline (approved):
i just had one thing regarding how the settings are being updated, once you take care of that i think this is good to merge.

christopappas:
Also a chance you might need to add these variables to test_settings in this repo

As for the middleware I think we decided not to go that route. Does that sounds right @brittneyexline ?

brittneyexline:
@muhammad-ammar ah yes - i missed that question. we decided not to use the middleware in favor of accessing the session cookie data another way, which @christopappas already implemented in edx-rbac.

enterprise_data/rules.py
muhammad-ammar: @brittneyexline @christopappas Regarding the above, [request_user_has_implicit_access_via_jwt](url) will tells that if a specific role is present in request for a user or not. In this case all the views uses `HasDataAPIDjangoGroupAccess` permission which means [ENTERPRISE_DATA_ADMIN_ROLE](https://github.com/edx/edx-enterprise-data/pull/110/files#diff-3bcb303d6a4b054f6c700234d4e0f8cf) maps to an `enterprise_admin` role, so if reques user has the `enterprise_admin` role than that user is allowed to access an endpoint, Is this check enough of we need to move all the logic present in [HasDataAPIDjangoGroupAccess](https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/permissions.py#L70) into `predicate`?

In short, I am trying understand how to decide what to put in a `rule predicate`?

brittneyexline: you can effectively ignore the HasDataAPIDjangoGroupAccess - this is the current authz check that we have that we want to get rid of. for the predicate, i would imagine that we structure it as one predicate check for implicit access, and one predicate for explicit access using https://github.com/edx/edx-rbac/blob/master/edx_rbac/utils.py#L90. then, we would register a permission check that maps to implicit_data_admin | explicit_data_admin, and use that permission for the view/viewset. does that make sense?

muhammad-ammar: Thanks. Sounds good. I will look more into this and discuss with you if there is any issue.

PR #111, opened 2019-03-20
django admin interface for role models

muhammad-ammar:

Add django admin interface for feature role models.

### DIFF
--- a/enterprise_data/admin/__init__.py
+++ b/enterprise_data/admin/__init__.py
@@ -1,0 +1,39 @@
+# -*- coding: utf-8 -*-
+"""
+Django admin integration for enterprise-data.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from edx_rbac.admin import UserRoleAssignmentAdmin, UserRoleAssignmentAdminForm
+
+from django.contrib import admin
+
+from enterprise_data.models import EnterpriseDataRoleAssignment
+
+
+class EnterpriseDataRoleAssignmentAdminForm(UserRoleAssignmentAdminForm):
+    """
+    Djaong admin form for EnterpriseDataRoleAssignmentAdmin.
+    """
+
+    class Meta:
+        """
+        Meta class for EnterpriseDataRoleAssignmentAdminForm.
+        """
+        model = EnterpriseDataRoleAssignment
+        fields = ('user', 'role')
+
+
+@admin.register(EnterpriseDataRoleAssignment)
+class EnterpriseDataRoleAssignmentAdmin(UserRoleAssignmentAdmin):
+    """
+    Django admin for EnterpriseDataRoleAssignment Model.
+    """
+
+    class Meta(object):
+        """
+        Meta class for EnterpriseDataRoleAssignmentAdmin.
+        """
+        model = EnterpriseDataRoleAssignment
+
+    form = EnterpriseDataRoleAssignmentAdminForm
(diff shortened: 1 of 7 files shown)


enterprise_data/admin/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+        model = EnterpriseDataFeatureRole
+
+    list_display = (
+        'name', 'description'
+    )
+
+
+class EnterpriseDataRoleAssignmentAdminForm(UserRoleAssignmentAdminForm):
+    """
+    Used for testing UserRoleAssignmentAdminForm.
```
christopappas: Should the docstring be changed here?

enterprise_data/admin/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+
+from edx_rbac.admin import UserRoleAssignmentAdmin, UserRoleAssignmentAdminForm
+
+from django.contrib import admin
+
+from enterprise_data.models import EnterpriseDataFeatureRole, EnterpriseDataRoleAssignment
+
+
+@admin.register(EnterpriseDataFeatureRole)
+class EnterpriseDataFeatureRoleAdmin(admin.ModelAdmin):
```
brittneyexline: i would advise removing this django admin for roles just because we should have tighter control over what roles exist in the system - i don't think there is a use case for non-engineers to add new roles whereas there is a need for non-engineers to assign roles to users.

brittneyexline (commented):
since this is the first time we are introducing the django admin site to be used by this service, it might be good to outline some testing instructions around it.

enterprise_data/admin/__init__.py
muhammad-ammar: great catch. 👍

enterprise_data/admin/__init__.py
muhammad-ammar: ok..will remove this.

Regarding `outline testing instructions`, should we add docs about how to test the admin interface or something else?

brittneyexline: no need for a separate doc - i just want to be certain there won't be some extra step for making sure the django admin screen is available when this is deployed to prod. so like what url did you hit on devstack to get to this? what do we expect it to be on stage and prod? just add this to the PR description. 
we are assuming this is straightforward, but i'd like to be sure that the assumption is correct.

brittneyexline (approved):
this seems ready to go, i just want to understand if anything unexpected is entailed for introducing django admin for the first time in the analytics data api service. it's just meant as a sanity check 😄

enterprise_data/admin/__init__.py
muhammad-ammar: @brittneyexline @christopappas @georgebabey  I have checked the admin interface in `edx-analytics-data-api` and I can see the `EnterpriseDataRoleAssignmentAdmin` there. But get issues when try to add a role assignment. There are two databases configured in local settings in `edx-analytics-data-api`. To make things work I need to

- Create a user in both the databases, user id must be same for both the users. In my case I have only one user.  
- Run the server using `DJANGO_DATABASE='analytics' python manage.py runserver`
- Than I will be able to create a role assignment record

Below is the error if things are not configured as mentioned above.
Screen Shot 2019-03-21 at 7 36 32 PM

Any idea about this issue?

muhammad-ammar:
@christopappas @brittneyexline Merging this. Will fix the django admin issue in `edx-analytics-data-api. Create [ticket](https://openedx.atlassian.net/browse/ENT-1718) for that.

PR #112, opened 2019-03-20
Cpappas/ent 1695

christopappas:

### DIFF
--- a/enterprise_data/api/v0/views.py
+++ b/enterprise_data/api/v0/views.py
@@ -7,6 +7,8 @@ from __future__ import absolute_import, unicode_literals
 from datetime import date, timedelta
 from logging import getLogger
 
+import waffle
+from edx_rbac.mixins import PermissionRequiredMixin
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 from edx_rest_framework_extensions.paginators import DefaultPagination
 from rest_framework import filters, viewsets
@@ -20,6 +22,7 @@ from django.db.models.functions import Coalesce
 from django.utils import timezone
 
 from enterprise_data.api.v0 import serializers
+from enterprise_data.constants import ROLE_BASED_ACCESS_CONTROL_SWITCH
 from enterprise_data.filters import (
     CONSENT_TRUE_OR_NOENROLL_Q,
     AuditEnrollmentsFilterBackend,
@@ -29,6 +32,28 @@ from enterprise_data.models import EnterpriseEnrollment, EnterpriseUser
 from enterprise_data.permissions import HasDataAPIDjangoGroupAccess
 
 LOGGER = getLogger(__name__)
+
+
+class EnterprisePermissionRequiredMixin(PermissionRequiredMixin):
+    """
+    Customized mixin for doing permission checks permissions.py and Django Rules.
+    """
+
+    def check_permissions(self, request):
+        """
+        Check permissions via permissions.py and Django Rules
+        """
+        # edx_rbac PermissionRequiredMixin overrides the following method in DRF:
+        #
+        #   https://github.com/encode/django-rest-framework/blob/master/rest_framework/views.py#L337
+        #
+        # We need to keep this DRF method logic intact for now so we can run
+        # the code in permissions.py. Specifically: HasDataAPIDjangoGroupAccess
+        if waffle.switch_is_active(ROLE_BASED_ACCESS_CONTROL_SWITCH):
+            # role based permission
+            super(EnterprisePermissionRequiredMixin, self).check_permissions(request)
+        else:
+            super(PermissionRequiredMixin, self).check_permissions(request)  # pylint: disable=bad-super-call
 
 
 def subtract_one_month(original_date):
@@ -42,13 +67,14 @@ def subtract_one_month(original_date):
     return one_month_earlier
 
 
-class EnterpriseViewSet(viewsets.ViewSet):
+class EnterpriseViewSet(EnterprisePermissionRequiredMixin, viewsets.ViewSet):
     """
     Base class for all Enterprise view sets.
     """
     authentication_classes = (JwtAuthentication,)
     pagination_class = DefaultPagination
     permission_classes = (HasDataAPIDjangoGroupAccess,)
+    permission_required = 'can_access_enterprise'
 
     def ensure_data_exists(self, request, data, error_message=None):
         """
--- a/enterprise_data/clients.py
+++ b/enterprise_data/clients.py
@@ -9,10 +9,11 @@ from edx_django_utils.cache import TieredCache
 from edx_rest_api_client.client import EdxRestApiClient
 from edx_rest_api_client.exceptions import HttpClientError, HttpServerError
 from rest_framework.exceptions import NotFound, ParseError
+from six.moves import urllib
 
 from django.conf import settings
 
-from enterprise_data.utils import get_cache_key
+from enterprise_data.utils import get_cache_key, update_session_with_enterprise_data
 
 DEFAULT_REPORTING_CACHE_TIMEOUT = 60 * 60 * 6  # 6 hours (Value is in seconds)
 LOGGER = logging.getLogger('enterprise_data')
@@ -62,9 +63,15 @@ class EnterpriseApiClient(EdxRestApiClient):
 
         return response['results'][0]
 
-    def get_with_access_to(self, user, enterprise_id):
+    def get_enterprise_customer(self, user, enterprise_id, with_access_to=True):
         """
-        Get the enterprises that this user has access to for the data api permission django group.
+        Get the enterprises that this user has access to.
+
+        with_access_to is True by default, which requires the data api permission django
+        group access.
+
+        If with_access_to is set to False, do not add the with_access_to check. The use
+        case of this would be if access to enterprise data was being controlled in the jwt.
         """
         cache_key = get_cache_key(
             resource='enterprise-customer',
@@ -76,20 +83,24 @@ class EnterpriseApiClient(EdxRestApiClient):
             return cached_response.value
 
         try:
-            querystring = {
-                'permissions': [self.ENTERPRISE_DATA_API_GROUP],
-                'enterprise_id': enterprise_id,
-            }
             endpoint = getattr(self, 'enterprise-customer')  # pylint: disable=literal-used-as-attribute
-            endpoint = endpoint.with_access_to
-            response = endpoint.get(**querystring)
+            if with_access_to is True:
+                querystring = {
+                    'enterprise_id': enterprise_id,
+                    'permissions': [self.ENTERPRISE_DATA_API_GROUP]
+                }
+                endpoint = endpoint.with_access_to
+                response = endpoint.get(**querystring)
+            else:
+                endpoint = urllib.parse.urljoin(endpoint.url(), enterprise_id)
+                response = endpoint.get()
         except (HttpClientError, HttpServerError) as exc:
-            LOGGER.warning("Unable to retrieve Enterprise Customer with_access_to details for user {}: {}"
+            LOGGER.warning("Unable to retrieve Enterprise Customer details for user {}: {}"
                            .format(user.username, exc))
             raise exc
 
         if response.get('results', None) is None:
-            raise NotFound('Unable to process Enterprise Customer with_access_to details for user {}, enterprise {}:'
+            raise NotFound('Unable to process Enterprise Customer details for user {}, enterprise {}:'
                            ' No Results Found'
                            .format(user.username, enterprise_id))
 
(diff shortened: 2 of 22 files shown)


muhammad-ammar:
@brittneyexline @christopappas This is ready to review. One test is failing and I am trying to understand why it is failing.

enterprise_data/api/v0/views.py
```diff
@@ -31,6 +32,25 @@
...
+        Check permissions via permissions.py and Django Rules
+        """
+        # edx_rbac PermissionRequiredMixin overrides the following method in DRF:
+        #
+        #   https://github.com/encode/django-rest-framework/blob/master/rest_framework/views.py#L337
+        #
+        # We need to keep this DRF method logic intact for now so we can run
+        # the code in permissions.py. Specifically: HasDataAPIDjangoGroupAccess
+        super(PermissionRequiredMixin, self).check_permissions(request)  # pylint: disable=bad-super-call
+        super(EnterprisePermissionRequiredMixin, self).check_permissions(request)
```
muhammad-ammar: @brittneyexline @christopappas Can we get rid of [HasDataAPIDjangoGroupAccess](https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/permissions.py#L70) permission? Isn't the role permissions are enough?

christopappas: So this is actually going to be the bit i think we will wrap in a waffle switch. when the switch is off, we'll want to use the HasDataAPIDjangoGroupAccess rules, and then when the switch is on, we'll use the new roles. For the purpose of blocking off HasDataAPIDjangoGroupAccess rules in a switch, we might have to refactor a slight bit to make sure the things we're adding to the session in HasDataAPIDjangoGroupAccess are still added

enterprise_data/rules.py
```diff
@@ -0,0 +1,48 @@
...
+    request_user_has_implicit_access_via_jwt,
+    user_has_access_via_database,
+)
+
+from enterprise_data.constants import ENTERPRISE_DATA_ADMIN_ROLE
+from enterprise_data.models import EnterpriseDataRoleAssignment
+
+
+@rules.predicate
+def is_request_user_has_implicit_access(*args, **kwargs):  # pylint: disable=unused-argument
```
christopappas: nit/personal preference, but i'd probably drop the `is_` in the name making it `request_user_has_implicit_access`

enterprise_data/rules.py
```diff
@@ -0,0 +1,48 @@
...
+    Returns:
+        boolean: whether the request user has access or not
+    """
+    request = get_request_or_stub()
+    decoded_jwt = get_decoded_jwt_from_request(request)
+    return request_user_has_implicit_access_via_jwt(decoded_jwt, ENTERPRISE_DATA_ADMIN_ROLE)
+
+
+@rules.predicate
+def is_request_user_has_explicit_access(*args, **kwargs):  # pylint: disable=unused-argument
```
christopappas: nit/personal preference again, maybe making this `request_user_has_explicit_access`

enterprise_data/tests/test_filters.py
```diff
@@ -7,13 +7,14 @@
...
-from rest_framework.test import APITestCase
+from rest_framework.test import APITransactionTestCase
 
 from enterprise_data.models import EnterpriseEnrollment
+from enterprise_data.tests.mixins import JWTTestMixin
 from enterprise_data.tests.test_utils import UserFactory
 
 
-class TestConsentGrantedFilterBackend(APITestCase):
+class TestConsentGrantedFilterBackend(JWTTestMixin, APITransactionTestCase):
```
christopappas: For my own learning, why did we switch over to APITransactionTestCase here? Were there failing tests after adding in the jwt mixing?

enterprise_data/tests/test_views.py
```diff
@@ -515,17 +524,47 @@ def test_no_page_querystring_skips_pagination(self):
...
+    )
+    @ddt.unpack
+    def test_permissions(self, implicit_perm, explicit_perm, status):  # pylint: disable=redefined-outer-name
+        """
+        Test that role base permissions works as expected.
+        """
+        system_wide_role = SYSTEM_ENTERPRISE_ADMIN_ROLE
+
+        if implicit_perm is False:
+            system_wide_role = 'blahblah'
```
christopappas: nitpick, maybe changing `blahblah` to something like `role_with_no_mapped_permissions` to make it a tad more readable

enterprise_data/tests/test_views.py
```diff
@@ -1043,36 +1084,59 @@ def test_no_page_querystring_skips_pagination(self):
...
```
christopappas: nitpick, same comment here about blahblah :)

enterprise_data/tests/test_views.py
```diff
@@ -1243,3 +1314,28 @@ def test_viewset_ordering(
...
```
christopappas: nitpick - and here~

christopappas (commented):
I added a few nitpicky things and a question, but overall I think this looks great Ammar! Thank you for adding those tests, as well as all the JWT stuff

I'm going to hopefully add the waffle switch logic in today as well, so stay tuned for that

christopappas:
@brittneyexline @muhammad-ammar -- i made one last refactor for the waffle switch work. specifically: I changed how the EnterpriseClient's `with_access_to` method works, so that when the waffle switch is on, we can simply get the enterprise API for enterprise customer data without worrying about the with_access_to check (because as I understand it that is something we want to drop with the rollout of role based authentication)

let me know what comments/concerns you might have!

thanks!

enterprise_data/clients.py
```diff
@@ -77,19 +83,20 @@ def get_with_access_to(self, user, enterprise_id):
...
             querystring = {
-                'permissions': [self.ENTERPRISE_DATA_API_GROUP],
                 'enterprise_id': enterprise_id,
             }
             endpoint = getattr(self, 'enterprise-customer')  # pylint: disable=literal-used-as-attribute
-            endpoint = endpoint.with_access_to
+            if with_access_to is True:
+                querystring.update({'permissions': [self.ENTERPRISE_DATA_API_GROUP]})
+                endpoint = endpoint.with_access_to
             response = endpoint.get(**querystring)
```
brittneyexline: does the regular enterprise customer endpoint filter to a specific enterprise by querystring? i thought it was part of the url path...

enterprise_data/clients.py
```diff
@@ -103,27 +110,26 @@ def get_with_access_to(self, user, enterprise_id):
...
-            Enterprise or None if unable to get or user is not associated with an enterprise
+            Enterprise Customer or None if unable to get enterprise information
         """
         enterprise_id = request.parser_context.get('kwargs', {}).get('enterprise_id', '')
         session = request.session
         enterprise_data = None
 
-        if 'enterprises_with_access' not in session or enterprise_id not in session['enterprises_with_access']:
-            enterprise_data = self.get_with_access_to(request.user, enterprise_id)
+        if 'enable_audit_enrollment' not in session or 'enforce_data_sharing_consent' not in session:
```
brittneyexline, line 128: i think we'll need to check if the enterprise id is in `session['enable_audit_enrollment']` and `session['enforce_data_sharing_consent']` based on how `update_session_with_enterprise_data works`: https://github.com/edx/edx-enterprise-data/blob/27fba89d848668a4a35cd250aa94868442ea2e58/enterprise_data/utils.py#L12

enterprise_data/api/v0/views.py
```diff
@@ -31,6 +34,28 @@
...
+        #
+        #   https://github.com/encode/django-rest-framework/blob/master/rest_framework/views.py#L337
+        #
+        # We need to keep this DRF method logic intact for now so we can run
+        # the code in permissions.py. Specifically: HasDataAPIDjangoGroupAccess
+        if waffle.switch_is_active(ROLE_BASED_ACCESS_CONTROL_SWITCH):
+            # role based permission
+            super(EnterprisePermissionRequiredMixin, self).check_permissions(request)
+        else:
+            super(PermissionRequiredMixin, self).check_permissions(request)  # pylint: disable=bad-super-call
```
brittneyexline, line 56: should we just return here instead of making the "bad" super call? if we want to skip this logic entirely, is it necessary to return the super class's logic?

enterprise_data/rules.py
```diff
@@ -0,0 +1,48 @@
...
+def request_user_has_implicit_access(*args, **kwargs):  # pylint: disable=unused-argument
+    """
+    Check that if request user has implicit access to `ENTERPRISE_DATA_ADMIN_ROLE` feature role.
+
+    Returns:
+        boolean: whether the request user has access or not
+    """
+    request = get_request_or_stub()
+    decoded_jwt = get_decoded_jwt_from_request(request)
+    return request_user_has_implicit_access_via_jwt(decoded_jwt, ENTERPRISE_DATA_ADMIN_ROLE)
```
brittneyexline, line 28: i am realizing now that we haven't addressed checking a matching enterprise for this role. it might be easier to do that in a future PR (we can take advantage of the fact that we're releasing dark right now).

this means for implicit check, we need to get the enterprise id associated with the request object and see if it matches the context for the role. for the explicit check, i think we might add a nullable enterprise id field and have a get_context method implemented that returns the enterprise id. then here, we will check if the enterprise id associated with the request object matches get_context on the model instance.

brittneyexline (commented):
thanks @christopappas and @muhammad-ammar for your hard work on this! there are a lot of moving parts, so i think we should make sure we can iterate safely, which means making sure 1) the waffle switch works and 2) doesn't break anything when it's off. i would like to see some testing instructions on devstack related to those things before we merge this particular PR. the specific asks around checking the user's enterprise can come as a follow up.

enterprise_data/api/v0/views.py
christopappas: `PermissionRequiredMixin` in edx-rbac is what allows us to put role based access on viewsets. However, when we add the mixin to the viewset, it overrides the original check_permissions method that exists in vanilla drf. What is means is, while this mixin is part of the viewset, the previous django permission logic we were using will never work.

To that end, we have the 2 super calls based on the waffle switch. The first of which will let the new role based access permission logic happen, the 2nd of which will force the old logic to run

brittneyexline: got it.

enterprise_data/rules.py
christopappas: `we might add a nullable enterprise id field and have a get_context method implemented that returns the enterprise id`

on which model? the roleAssignment?

brittneyexline: correct.

enterprise_data/clients.py
christopappas: @brittneyexline i think this is good to go with my latest commit

muhammad-ammar:
@brittneyexline @christopappas I have tested the following 2 scenarios on devstack as per [this comment](https://edx-internal.slack.com/archives/CDXAQ8G02/p1553296830029700). Let me know if something is missing in testing?

1. As a staff user, confirm you can log in to the portal and see all customers available to you (without getting any 403s or other unexpected errors)
2. As a non-staff admin user, confirm you can log in to the portal and see ONLY the enterprise you are associated with

**Repos**

| repo        | git HEAD           |
| ------------- |:-------------:|
| ecommerce     | latest master |
| edx-portal      | latest master      |
| edx-analytics-data-api | latest master      |
| edx-enterprise-data | cpappas/ENT-1695     |

**Waffle Switch in edx-enterprise-data:** OFF

**Tested Steps:**

1. Created 2 enterprises
    a. batman
    b. riddler
2. Created two users --- both have `enterprise_data_api_access` group
    a. edx@example.com as staff admin
    b. edx2@example.com as non-staff admin
3. Linked `edx2@example.com` with `riddler` enterprise
4. Created `enterprise_user` and `enterprise_enrollment` records in `edx-enterprise-data` using management commands
4. Logged in with `edx@example.com` on portal
    a. I am able to see both the enterprises and learners associated with each
5. Logged in with `edx2@example.com` on portal
    a. I am only able to see the learners associated with `riddler` enterprise

enterprise_data/clients.py
muhammad-ammar: I have updated the logic in this method. `get_enterprise_and_update_session` is only used by `enterprise_data/filters.py` and they have custom logic there to check `enterprise_id`

christopappas:
This just needs 👍 i believe. I've rebased and bumped the version number

muhammad-ammar (approved):

enterprise_data/clients.py
brittneyexline: even with the logic in the filters, i think this will be a problem for edx staff users who have access to all enterprises, so this data does need to be checked on a per enterprise basis in that case, because otherwise we'll be assuming the data for an enterprise is there when it might not be. that being said, i am ok if we address this in another PR since it only affects logic with the flag off.

brittneyexline (approved):

enterprise_data/clients.py
muhammad-ammar: Unfortunately I am not able to understand what exactly is the issue here? :( Can you please explain? Also `get_enterprise_and_update_session` only executes when flag is active/on, so should this be addressed now or in a separate PR?

@christopappas Can you please look into this ^^

christopappas: Brittney, I remember you talking last week about a Null value existing for staff users. Aka, if they have a role assignment that is not associated with any particular enterprise, then it would be considered they have access to all of them. Is that related to this?

Also, if that were true, that this call would have a problem `enterprise_data = self.get_enterprise_customer(request.user, enterprise_id, with_access_to=False)` right? cause enterprise_id would be None?

brittneyexline: so we should be determining which enterprise_id to use based on the request, not the user, because we are asking about permission to the requested resource. so enterprise_id should never be determined by the user, but should be compared to the user's enterprise depending on their role.
staff users - or enterprise_openedx_operators as we have defined the role - may or may not have an enterprise attached, but it shouldn't matter because of the role. in that case, no check against the user's enterprise should be performed.
with regards to this logic, it is responsible for populating certain data about a given enterprise which affects what data is served to the user. the structure of the session storage for a user that has access to multiple enterprises (ie enterprise_openedx_operators) should be something like:
```
{ 'enable_audit_enrollment': { 'enterprise1': true, 'enterprise2': false }, 'enforce_data_sharing_consent': { 'enterprise1': true, 'enterprise2': false }
```

the issue with the logic in this line is that let's say the staff user goes and visits enterprise1's page. this is the first time they come through, so the keys don't exist in the session, and we make the call to get enterprise1's data. this behaves as expected. the user visits enterprise2's page, and the keys exist in the session, but enterprise2's data has not been populated, so the request will not behave as expected.

to fix: the line `if 'enable_audit_enrollment' not in session or 'enforce_data_sharing_consent' not in session:` will need to be changed to: `(if 'enable_audit_enrollment' not in session or enterprise_id not in session['enable_audit_enrollment']) or ('enforce_data_sharing_consent' not in session or enterprise_id not in session['enforce_data_sharing_consent']):`

muhammad-ammar: @brittneyexline Thanks a lot for the detailed explanation. I am merging this PR and will add the above mentioned changes in https://github.com/edx/edx-enterprise-data/pull/113

PR #113, opened 2019-03-26
improve role permission checks

muhammad-ammar:

**Description:** This PR adds context checking in implicit and explicit permission checks.

**Testing**

Done some testing previously with waffle switch OFF, Please see https://github.com/edx/edx-enterprise-data/pull/112#issuecomment-476555848

Waffle Switch OFF:

- staff users with the enterprise_data_api_access group who are may or may not be linked to an enterprise get access.
- non-staff users linked to the same enterprise as the request object who have the enterprise_data_api_access group get access.
- all other users are denied (staff without the group, matching enterprise users without the group, non matching enterprise users with the group)

Waffle Switch ON:

- users with the system wide role in their jwt and a matching enterprise get access.
- users with a feature specific role assignment with a matching enterprise get access.
- users with an operator role get access.
- all other users are denied (this includes staff)

## 

**Testing Details**

Test Enterprises

-  name = batman, uuid = 6ae013d4-c5c4-474d-8da9-0e559b2448e2
-  name = riddler, uuid = 885f4d97-5a21-4e8a-8723-a434bc527e74

##

##

* **Case:** With waffle switch OFF, staff users with the enterprise_data_api_access group who are may or may not be linked to an enterprise get access.
    * user100@example.com
        * staff = True
        * has enterprise_data_api_access
        * linked to enterprise batman
    *  user100@example.com
        * staff = True
        * has enterprise_data_api_access
        * not linked to any enterprise
* **Result:** HTTP_200_OK and user100@example.com is able to access all the enterprises and data with in them

##

* **Case:** With waffle switch OFF, non-staff users linked to the same enterprise as the request object who have the enterprise_data_api_access group get access.
    * user101@example.com              
        * staff = False
        * has enterprise_data_api_access
        * linked to enterprise riddler
* **Result:** HTTP_200_OK
* **NOTE:** 
     - user100@example.com is able to access the riddler enterprise only
     - gets HTTP_401 if try to access an endpoint with any enterprise id other than riddler

##

* **Case:** With waffle switch OFF, all other users are denied (staff without the group, matching enterprise users without the group, non matching enterprise users with the group)
    * user102@example.com         
        * staff = True
        * has no enterprise_data_api_access
        * not linked to any enterprise
   * **Result:** HTTP_403_FORBIDDEN

    * user103@example.com 
        * staff = False
        * has no enterprise_data_api_access
        * linked with riddler enterprise
    * **Result:** HTTP_403_FORBIDDEN

##

### Waffle Switch ON

* Create `SystemWideEnterpriseUserRoleAssignment` for a user for `enterprise_admin` role
* Add `SYSTEM_WIDE_ROLE_CLASSES` set to `['enterprise.SystemWideEnterpriseUserRoleAssignment']` in edx-platform
* Add `crum.CurrentRequestUserMiddleware `in `MIDDLEWARE_CLASSES` in edx-analytics-data-api
* Add `SYSTEM_TO_FEATURE_ROLE_MAPPING` in edx-analytics-data-api
* Link the user with an enterprise so that correct context is attached to role — do this in lms enterprise django admin

Dependency: https://github.com/edx/edx-analytics-data-api/pull/257

**NOTE:** user must have `enterprise_data_api_access` group for testing
            * This should not be required but currently we are making calls to `with_access_to` endpoint and that endpoint requires `enterprise_data_api_access` group to be present otherwise we will get `HTTP_403_FORBIDDEN`
            * I have tested with [with_access_to](https://github.com/edx/edx-portal/blob/master/src/data/services/LmsApiService.js#L29) removed calls from edx-portal and it works

##

* **Case:** With waffle switch ON, users with the system wide role in their jwt and a matching enterprise get access.
   * user200@example.com
        * has `enterprise_data_api_access` group
        * linked with `batman` enterprise
        * has `enterprise_admin` role assignment
        * role data is `enterprise_admin:6ae013d4-c5c4-474d-8da9-0e559b2448e2`
    * **Result:** HTTP_200_OK
##

* **Case:** With waffle switch ON, users with a feature specific role assignment with a matching enterprise get access.
   * user201@example.com 
        * has `enterprise_data_api_access` group
        * linked with `riddler` enterprise
        * create `EnterpriseDataRoleAssignment` record with `enterprise_admin` role
        * role data is `enterprise_admin:885f4d975a214e8a8723a434bc527e74`
    * **Result:** HTTP_200_OK

##

* **Case:** With waffle switch ON, users with an operator role get access.
   * user203@example.com 
        * has `enterprise_data_api_access` group
        * has `enterprise_openedx_operator` role assignment
        * not linked with any enterprise
        * role data in jwt is `enterprise_openedx_operator:*`
    * **Result:** HTTP_200_OK

##

* **Case:** With waffle switch ON, all other users are denied (this includes staff)
   * edx@example.com 
        * has `enterprise_data_api_access` group
        * `superuser` and `staff`
        * no role assignment
        * linked with `batman` enterprise
    * **Result:** HTTP_403_FORBIDDEN
   * **NOTE:** User is able to see all the enterprises but gets error when try to access any enterprise


muhammad-ammar:
@christopappas @brittneyexline I have started working on enterprise id check in implicit and explicit permission checks. Is this the only thing we need to add for the follow up from ENT-1695?

enterprise_data/models.py
```diff
@@ -130,6 +131,40 @@ def __repr__(self):
...
+class EnterpriseRoleAssignmentContextMixin(object):
+    """
+    Mixin for RoleAssignment models related to enterprises.
+    """
+
+    @property
+    def enterprise_customer_uuid(self):
+        """Get the enterprise customer uuid linked to the user."""
+        try:
+            enterprise_user = EnterpriseUser.objects.get(user_email=self.user.email)
```
muhammad-ammar: @christopappas @brittneyexline I am not sure if this is best way to get user's enterprise_id, as per my understanding data in `EnterpriseUser` model is for analytics purposes, so should we make a call to lms and get enterprise_id for the user or the above implementation is ok?

enterprise_data/rules.py
```diff
@@ -11,10 +11,30 @@
...
 
+def has_correct_role_assignment_context(request):
+    """
+    Check if request has correct role assignement context.
+    """
+    __, __, kwargs = resolve(request.path)
+    enterprise_id = kwargs.get('enterprise_id', '')
+
+    try:
+        role_assignement = EnterpriseDataRoleAssignment.objects.get(
```
christopappas: nitpick but "role_assignement" should be "role_assignment" here and below

enterprise_data/models.py
```diff
@@ -130,6 +131,40 @@ def __repr__(self):
...
+            return None
+        except MultipleObjectsReturned:
+            LOGGER.warning(
+                'User {} is linked to multiple enterprises, which is not yet supported!'.format(self.user.id)
+            )
+            return None
+
+        return str(enterprise_user.enterprise_id)
+
+    def get_context(self):
```
christopappas: In edx-enterprise I believe we simply return the uuid, and not a dict

Question for all: should we try to keep these get_context methods across our systems similar where we can (provided of course nothing else is required specifically)?

enterprise_data/models.py
christopappas: I think I'll need to defer to Brittney's thoughts, but my take is:

it seems like it would be fine to use EnterpriseUser's `enterprise_id` as I believe that value is what ultimately comes from the LMS; HOWEVER, because data is piped in daily, if ever there was a change to the value in the LMS, it would take a day for the data api to be updated, which could result in loss of access for people for a day. Specifically because the uuid from the request which we're checking in `has_correct_role_assignment_context` would not match the uuid of what was returned by `role_assignement.get_context().get('enterprise_customer_id')` for a day)

christopappas: So I guess if we're okay with this lag time, then we probably don't need to make an lms call

^ Does this sound correct? :)

brittneyexline: i guess my original thought was that it would be set explicitly on the role assignment as a new field, ie you are explicitly granting this user a role for this enterprise. i hadn't considered that we might infer their enterprise association. 
i'm inclined to say that we might want the ability to explicitly state the enterprise with the feature specific role assignment in this case, ie maybe we'll need to give an internal user who doesn't normally have broad enterprise access the ability to see the data for an academic enterprise that they manage. i think adding the enterprise id as a field to the role assignment object will give us that flexibility.

enterprise_data/models.py
brittneyexline: i think until we have a need to return more structured data, we should just return the uuid here. ideally we wouldn't want to have very complicated context data.

enterprise_data/rules.py
```diff
@@ -11,10 +11,30 @@
     user_has_access_via_database,
 )
 
+from django.urls import resolve
+
 from enterprise_data.constants import ENTERPRISE_DATA_ADMIN_ROLE
 from enterprise_data.models import EnterpriseDataRoleAssignment
 
 
+def has_correct_role_assignment_context(request):
```
brittneyexline: this logic will only work for the explicit check, since it is looking at the feature specific role assignment.

enterprise_data/rules.py
```diff
@@ -24,8 +44,14 @@ def request_user_has_implicit_access(*args, **kwargs):  # pylint: disable=unused
...
     request = get_request_or_stub()
+
     decoded_jwt = get_decoded_jwt_from_request(request)
-    return request_user_has_implicit_access_via_jwt(decoded_jwt, ENTERPRISE_DATA_ADMIN_ROLE)
+    implicit_access = request_user_has_implicit_access_via_jwt(decoded_jwt, ENTERPRISE_DATA_ADMIN_ROLE)
+
+    if not implicit_access:
+        return False
+
+    return has_correct_role_assignment_context(request)
```
brittneyexline: for the implicit check, we need to get the user's enterprise association from the jwt token, not from the feature specific role assignment.

brittneyexline (commented):
happy to see so much progress, keep it up!
to answer your question about anything else needing to be addressed from the previous PR, it would be making sure for staff users in a single session visiting multiple enterprises, that we are correctly populating the session with the specific enterprises data after the first enterprise they look at (see: https://github.com/edx/edx-enterprise-data/pull/112#discussion_r269688472).

enterprise_data/models.py
muhammad-ammar: Thanks brittney, things are clear and this is much more easy.

enterprise_data/rules.py
```diff
@@ -11,10 +11,51 @@
...
+    """
+    Check if request has correct role assignement context.
+    """
+    __, __, kwargs = resolve(request.path)
+    enterprise_id_in_request = kwargs.get('enterprise_id')
+
+    jwt_roles_claim = decoded_jwt.get('roles', [])
+
+    for role_data in jwt_roles_claim:
+        role_in_jwt, enterprise_id_in_jwt = role_data.split(':')
```
muhammad-ammar: **Note to Reviewers:** Other than the `enterprise_openedx_operator` role, should I expect that there will be no context in role data?

muhammad-ammar:
@brittneyexline @christopappas Addressed the feedback. Please continue the review.

muhammad-ammar:
@brittneyexline @christopappas I have made the changes as per https://github.com/edx/edx-enterprise-data/pull/112#discussion_r270155746 Please have a look

enterprise_data/models.py
```diff
@@ -161,6 +161,13 @@ class EnterpriseDataRoleAssignment(UserRoleAssignment):
 
     role_class = EnterpriseDataFeatureRole
     user = models.ForeignKey(settings.AUTH_USER_MODEL, db_index=True, on_delete=models.CASCADE)
+    enterprise_id = models.UUIDField(blank=True, null=True, verbose_name='Enterprise Customer UUID')
```
christopappas: Now that Saleem's new django app for the roles and role assignments has been merged, I think you can rebase/move this over to the model over there

enterprise_data/models.py
```diff
@@ -161,6 +161,13 @@ class EnterpriseDataRoleAssignment(UserRoleAssignment):
...
+
+    def get_context(self):
+        """
+        Return the context for this role assignment class.
+        """
+        return self.enterprise_id and str(self.enterprise_id)
```
christopappas: This would return a bool, right? Should instead we expect get_context() to return some uuid or None, and then let the caller do something based on the result?

If there's a good motivation for the bool, that's fine, but I was under the impression that get_context methods would  typically return some kind of non-bool value (like a uuid)

enterprise_data/models.py
muhammad-ammar: Moved the migration to new app

enterprise_data/models.py
muhammad-ammar: Updated the logic and now it will return `None`  or enterprise id

muhammad-ammar:
@christopappas Addressed your feedback.

enterprise_data/tests/test_views.py
```diff
@@ -530,26 +531,86 @@ def test_no_page_querystring_skips_pagination(self):
...
 
     @ddt.data(
-        {'implicit_perm': False, 'explicit_perm': False, 'status': status.HTTP_403_FORBIDDEN},
-        {'implicit_perm': False, 'explicit_perm': True, 'status': status.HTTP_200_OK},
-        {'implicit_perm': True, 'explicit_perm': False, 'status': status.HTTP_200_OK},
+        {
+            'implicit': False,
+            'explicit': False,
+            'jwt_enterprise_id': None,
+            'context_enterprise_id': None,
```
christopappas, line 538: the bool examples here for context_enterprise_id need to updated to some string i believe (based on the other changes to get_context()

christopappas: nevermind, i misread the test

enterprise_data_roles/rules.py
```diff
@@ -11,10 +11,51 @@
...
+
+    for role_data in jwt_roles_claim:
+        role_in_jwt, enterprise_id_in_jwt = role_data.split(':')
+        if role_in_jwt == system_wide_role_name:
+            return enterprise_id_in_jwt == enterprise_id_in_request
+
+    return False
+
+
+def has_correct_context_for_explicit_access(request):
```
brittneyexline, line 37: nit: since this function is only checking for a specific role (enterprise_data_admin), the name should reflect that.

brittneyexline (commented):
seems like it might make sense to put context checks as utility functions in edx-rbac, but i don't think that should block merging.
also, here are the test cases i gave irfan for ecommerce, they should largely be the same:
Waffle switch off:
* staff users with the enterprise_data_api_access group who are may or may not be linked to an enterprise get access.
* non-staff users linked to the same enterprise as the request object who have the enterprise_data_api_access group get access.
* all other users are denied (staff without the group, matching enterprise users without the group, non matching enterprise users with the group)

Waffle switch on:
* users with the system wide role in their jwt and a matching enterprise get access.
* users with a feature specific role assignment with a matching enterprise get access.
* all other users are denied (this includes staff)

brittneyexline (approved):

muhammad-ammar:
@brittneyexline @christopappas I am going to merge this. I have completed testing with waffle switch OFF. Please see PR description. Let me know if some test case is missing.

I am currently doing testing with waffle switch ON and add all the fixes on https://github.com/edx/edx-enterprise-data/pull/116

PR #116, opened 2019-04-01
add enterprise-openedx-operator permission checks

muhammad-ammar:
[ENT-1738](https://openedx.atlassian.net/browse/ENT-1711)

Description: Add permission checks for `enterprise_openedx_operator` role and update the logic for existing implicit and explicit access checks.

Dependencies:
1. https://github.com/edx/edx-rbac/pull/16
2. https://github.com/edx/edx-enterprise/pull/450

**Testing**

Done some testing previously with waffle switch OFF, Please see https://github.com/edx/edx-enterprise-data/pull/112#issuecomment-476555848

Waffle Switch OFF:

- staff users with the enterprise_data_api_access group who are may or may not be linked to an enterprise get access.
- non-staff users linked to the same enterprise as the request object who have the enterprise_data_api_access group get access.
- all other users are denied (staff without the group, matching enterprise users without the group, non matching enterprise users with the group)

Waffle Switch ON:

- users with the system wide role in their jwt and a matching enterprise get access.
- users with a feature specific role assignment with a matching enterprise get access.
- users with an operator role get access.
- all other users are denied (this includes staff)

## 

**Testing Details**

Test Enterprises

-  name = batman, uuid = 6ae013d4-c5c4-474d-8da9-0e559b2448e2
-  name = riddler, uuid = 885f4d97-5a21-4e8a-8723-a434bc527e74

##

##

* **Case:** With waffle switch OFF, staff users with the enterprise_data_api_access group who are may or may not be linked to an enterprise get access.
    * user100@example.com
        * staff = True
        * has enterprise_data_api_access
        * linked to enterprise batman
    *  user100@example.com
        * staff = True
        * has enterprise_data_api_access
        * not linked to any enterprise
* **Result:** HTTP_200_OK and user100@example.com is able to access all the enterprises and data with in them

##

* **Case:** With waffle switch OFF, non-staff users linked to the same enterprise as the request object who have the enterprise_data_api_access group get access.
    * user101@example.com              
        * staff = False
        * has enterprise_data_api_access
        * linked to enterprise riddler
* **Result:** HTTP_200_OK
* **NOTE:** 
     - user100@example.com is able to access the riddler enterprise only
     - gets HTTP_401 if try to access an endpoint with any enterprise id other than riddler

##

* **Case:** With waffle switch OFF, all other users are denied (staff without the group, matching enterprise users without the group, non matching enterprise users with the group)
    * user102@example.com         
        * staff = True
        * has no enterprise_data_api_access
        * not linked to any enterprise
   * **Result:** HTTP_403_FORBIDDEN

    * user103@example.com 
        * staff = False
        * has no enterprise_data_api_access
        * linked with riddler enterprise
    * **Result:** HTTP_403_FORBIDDEN

##

### Waffle Switch ON

* Create `SystemWideEnterpriseUserRoleAssignment` for a user for `enterprise_admin` role
* Add `SYSTEM_WIDE_ROLE_CLASSES` set to `['enterprise.SystemWideEnterpriseUserRoleAssignment']` in edx-platform
* Add `crum.CurrentRequestUserMiddleware `in `MIDDLEWARE_CLASSES` in edx-analytics-data-api
* Add `SYSTEM_TO_FEATURE_ROLE_MAPPING` in edx-analytics-data-api
* Link the user with an enterprise so that correct context is attached to role — do this in lms enterprise django admin

Dependency: https://github.com/edx/edx-analytics-data-api/pull/257

**NOTE:** user must have `enterprise_data_api_access` group for testing
            * This should not be required but currently we are making calls to `with_access_to` endpoint and that endpoint requires `enterprise_data_api_access` group to be present otherwise we will get `HTTP_403_FORBIDDEN`
            * I have tested with [with_access_to](https://github.com/edx/edx-portal/blob/master/src/data/services/LmsApiService.js#L29) removed calls from edx-portal and it works

##

* **Case:** With waffle switch ON, users with the system wide role in their jwt and a matching enterprise get access.
   * user200@example.com
        * has `enterprise_data_api_access` group
        * linked with `batman` enterprise
        * has `enterprise_admin` role assignment
        * role data is `enterprise_admin:6ae013d4-c5c4-474d-8da9-0e559b2448e2`
    * **Result:** HTTP_200_OK
##

* **Case:** With waffle switch ON, users with a feature specific role assignment with a matching enterprise get access.
   * user201@example.com 
        * has `enterprise_data_api_access` group
        * linked with `riddler` enterprise
        * create `EnterpriseDataRoleAssignment` record with `enterprise_admin` role
        * role data is `enterprise_admin:885f4d975a214e8a8723a434bc527e74`
    * **Result:** HTTP_200_OK

##

* **Case:** With waffle switch ON, users with an operator role get access.
   * user203@example.com 
        * has `enterprise_data_api_access` group
        * has `enterprise_openedx_operator` role assignment
        * not linked with any enterprise
        * role data in jwt is `enterprise_openedx_operator:*`
    * **Result:** HTTP_200_OK

##

* **Case:** With waffle switch ON, all other users are denied (this includes staff)
   * edx@example.com 
        * has `enterprise_data_api_access` group
        * `superuser` and `staff`
        * no role assignment
        * linked with `batman` enterprise
    * **Result:** HTTP_403_FORBIDDEN
   * **NOTE:** User is able to see all the enterprises but gets error when try to access any enterprise

### DIFF
--- a/enterprise_data/tests/mixins.py
+++ b/enterprise_data/tests/mixins.py
@@ -17,11 +17,13 @@ class JWTTestMixin(object):
         """
         Set jwt token in cookies
         """
+        role_data = '{system_wide_role}'.format(system_wide_role=system_wide_role)
+        if context is not None:
+            role_data += ':{context}'.format(context=context)
+
         payload = generate_unversioned_payload(self.user)
         payload.update({
-            'roles': [
-                '{system_wide_role}:{context}'.format(system_wide_role=system_wide_role, context=context)
-            ]
+            'roles': [role_data]
         })
         jwt_token = generate_jwt_token(payload)
 
--- a/enterprise_data_roles/constants.py
+++ b/enterprise_data_roles/constants.py
@@ -6,6 +6,10 @@ from __future__ import unicode_literals
 
 ENTERPRISE_DATA_ADMIN_ROLE = 'enterprise_data_admin'
 SYSTEM_ENTERPRISE_ADMIN_ROLE = 'enterprise_admin'
+SYSTEM_ENTERPRISE_OPERATOR_ROLE = 'enterprise_openedx_operator'
 
 # switch is used to enable/disable role based access control
 ROLE_BASED_ACCESS_CONTROL_SWITCH = 'role_based_access_control'
+
+# context to give access to all resources
+ALL_ACCESS_CONTEXT = '*'
--- a/enterprise_data_roles/rules.py
+++ b/enterprise_data_roles/rules.py
@@ -4,56 +4,13 @@ Rules needed to restrict access to the enterprise data api.
 from __future__ import absolute_import, unicode_literals
 
 import rules
-from edx_rbac.utils import (
-    get_decoded_jwt_from_request,
-    get_request_or_stub,
-    request_user_has_implicit_access_via_jwt,
-    user_has_access_via_database,
-)
+from edx_rbac.utils import get_request_or_stub, request_user_has_implicit_access_via_jwt, user_has_access_via_database
+from edx_rest_framework_extensions.auth.jwt.cookies import get_decoded_jwt as get_decoded_jwt_from_jwt_cookie
 
 from django.urls import resolve
 
-from enterprise_data_roles.constants import ENTERPRISE_DATA_ADMIN_ROLE, SYSTEM_ENTERPRISE_ADMIN_ROLE
+from enterprise_data_roles.constants import ENTERPRISE_DATA_ADMIN_ROLE
 from enterprise_data_roles.models import EnterpriseDataRoleAssignment
-
-
-def has_correct_context_for_implicit_access(request, decoded_jwt, system_wide_role_name):
-    """
-    Check if request has correct role assignement context.
-    """
-    __, __, kwargs = resolve(request.path)
-    enterprise_id_in_request = kwargs.get('enterprise_id')
-
-    jwt_roles_claim = decoded_jwt.get('roles', [])
-
-    for role_data in jwt_roles_claim:
-        role_in_jwt, enterprise_id_in_jwt = role_data.split(':')
-        if role_in_jwt == system_wide_role_name:
-            return enterprise_id_in_jwt == enterprise_id_in_request
-
-    return False
-
-
-def has_correct_context_for_explicit_access(request):
-    """
-    Check if request has correct role assignement context.
-    """
-    __, __, kwargs = resolve(request.path)
-    enterprise_id_in_request = kwargs.get('enterprise_id')
-
-    try:
-        role_assignment = EnterpriseDataRoleAssignment.objects.get(
-            user=request.user,
-            role__name=ENTERPRISE_DATA_ADMIN_ROLE
-        )
-    except EnterpriseDataRoleAssignment.DoesNotExist:
-        return False
-
-    # if there is no enterprise_id set than user is allowed
-    if role_assignment.get_context() is None:
-        return True
-
-    return role_assignment.get_context() == enterprise_id_in_request
 
 
 @rules.predicate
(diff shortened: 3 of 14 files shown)


christopappas (approved):
LGTM 👍

enterprise_data_roles/rules.py
```diff
@@ -27,9 +31,11 @@ def has_correct_context_for_implicit_access(request, decoded_jwt, system_wide_ro
     jwt_roles_claim = decoded_jwt.get('roles', [])
 
     for role_data in jwt_roles_claim:
-        role_in_jwt, enterprise_id_in_jwt = role_data.split(':')
+        role_in_jwt, __, enterprise_id_in_jwt = role_data.partition(':')
```
brittneyexline: we could probably refactor this check now to use the context checking functionality just added in edx-rbac

muhammad-ammar: will do in https://github.com/edx/edx-enterprise-data/pull/113 or in a separate future PR

enterprise_data_roles/constants.py
```diff
@@ -6,6 +6,7 @@
 
 ENTERPRISE_DATA_ADMIN_ROLE = 'enterprise_data_admin'
 SYSTEM_ENTERPRISE_ADMIN_ROLE = 'enterprise_admin'
+ENTERPRISE_OPERATOR_ROLE = 'enterprise_openedx_operator'
```
mushtaqak: Should this be renamed to `SYSTEM_ENTERPRISE_OPERATOR_ROLE` to maintain consistency?

enterprise_data/tests/mixins.py
```diff
@@ -17,11 +17,13 @@ def set_jwt_cookie(self, system_wide_role='enterprise_admin', context='some_cont
         """
         Set jwt token in cookies
         """
+        role_data = '{system_wide_role}'.format(system_wide_role=system_wide_role)
```
mushtaqak, line 20: an extra nit: can we use constant in method argument for `enterprise_admin`?

muhammad-ammar:
@brittneyexline Can we merge this?

enterprise_data_roles/rules.py
```diff
@@ -56,6 +36,25 @@ def has_correct_context_for_explicit_access(request):
     return role_assignment.get_context() == enterprise_id_in_request
 
 
+@rules.predicate
+def request_user_has_operator_access(*args, **kwargs):  # pylint: disable=unused-argument
```
brittneyexline: i'm hoping we can greatly simplify this predicate logic if we can get the rbac utils right. i think we'll want to implement the wildcard idea to help with this. the predicates here shouldn't have to dig into the jwt themselves to determine access - that is the whole purpose of the rbac package.

enterprise_data_roles/utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+import crum
+from six.moves.urllib.parse import urlparse  # pylint: disable=import-error
+
+from django.conf import settings
+from django.test.client import RequestFactory
+
+
+# Taken from openedx/core/lib/request_utils.py
+def get_request_or_stub():
```
brittneyexline: i'm pretty sure we have this function in edx-rbac, so there shouldn't be a need to copy, right? if you do need some adjustments, i'd rather see the edx-rbac function modified to meet your needs.

christopappas: agreed. this is where it lives in edx-rbac (with the note that i might live in drf extensions in the future) - https://github.com/edx/edx-rbac/blob/master/edx_rbac/utils.py#L17

christopappas (changes requested):
I believe we need to replace the get_request_or_stub function unless there is a need for it to be here as well as edx-rbac

enterprise_data_roles/utils.py
muhammad-ammar: I added this function here because of https://edx-internal.slack.com/archives/CDXAQ8G02/p1554212699018200

christopappas: ah, well, he's not going to delete that on us, he's asked us to remove it before we're done iterating on edx-rbac, so i think it's fine to leave the call from edx-rbac for now. additionally, we wouldn't just want to paste it here. we'd want to move the stuff in edx-rbac in question to edx-drf-extensions and import from there

so yeah, i think it's fine to keep the import from edx-rbac for now, because we are the ones in charge of moving that code over to edx-drf-extensions

i'll file a ticket to refactor edx-rbac once things are more stabile

muhammad-ammar: removed the new util and back to rbac get_request_or_stub

christopappas (approved):

enterprise_data_roles/rules.py
```diff
@@ -4,56 +4,27 @@
...
-    if role_assignment.get_context() is None:
-        return True
-
-    return role_assignment.get_context() == enterprise_id_in_request
+    request = get_request_or_stub()
+    decoded_jwt = get_decoded_jwt_from_jwt_cookie(request)
+    return request_user_has_implicit_access_via_jwt(
+        decoded_jwt,
+        ENTERPRISE_DATA_ADMIN_ROLE,
+        ALL_ACCESS_CONTEXT
```
brittneyexline: we should pass in the request enterprise id here instead.

brittneyexline (approved):
just one thing that needs to be changed before merging!

PR #123, opened 2019-04-22
Use get_decoded_jwt_from_request from edx-rbac

saleem-latif:
**Description:**
In this PR we are switching to use `get_decoded_jwt_from_request` from edx-rbac instead of the one present in `edx-drf-extensions`.


mushtaqak (approved):
LGTM-take-2 👍

irfanuddinahmad (approved):
LGTM!

PR #132, opened 2019-05-09
remove rbac switch

muhammad-ammar:
**Description:** This will remove the role base access control switch from db.

**JIRA:** [ENT-1885](https://openedx.atlassian.net/browse/ENT-1885)

**Problem:** [ROLE_BASED_ACCESS_CONTROL_SWITCH](https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data_roles/constants.py#L12) is defined as constant and is used in all the db migrations and if we remove this than travis builds will fail because database is created from a clean state and all the migrations are applied from first to last and migrations will fail when they are unable to find the constant. So it is not possible to remove this constant without manually modifying all the migrations.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,10 @@ Unreleased
 ----------
 
 =======
+[1.2.8] - 2019-05-17
+--------------------
+* Remove RBAC switch from DB.
+
 [1.2.7] - 2019-05-13
 --------------------
 * Replace edx_rbac.utils.get_decoded_jwt_from_request with edx_rest_framework_extensions.auth.jwt.cookies.get_decoded_jwt.
--- a/enterprise_data/__init__.py
+++ b/enterprise_data/__init__.py
@@ -4,6 +4,6 @@ Enterprise data api application. This Django app exposes API endpoints used by e
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "1.2.7"
+__version__ = "1.2.8"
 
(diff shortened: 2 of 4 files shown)


christopappas:
You bring up a really good point, Ammar

Do we think it's messy to remove all references to this after we merge this and remove the switch from the DB?

christopappas (approved):
Question about next steps, but I think this PR itself looks good!

christopappas:
@muhammad-ammar I think leaving in the constant variable is probably the easiest for now. As far as cruft goes, it's a very minor detail. Maybe we can add a comment by the constant variable saying what it was used for

The alternative would be to change the migrations by hand, or, to squash the migrations (therefore removing from the migrations any of the migrations that including this constant value). I think squashing is the "better" approach, although it would require a more attention to make sure we don't break people's checkouts and production

Given that migrations take just a few seconds to run, I think leaving the variable is a fine path forward. If we find migration time to be taking too long, we can revisit this

muhammad-ammar:
@christopappas Most of our rbac migrations are data migrations which can not be squashed as [per](https://docs.djangoproject.com/en/1.11/topics/migrations/#squashing-migrations) so I think leaving the constant is good for now.

christopappas:
Sounds good to me! Thanks @muhammad-ammar :D

mushtaqak (approved):
Looks good 👍

brittneyexline (approved):

PR #133, opened 2019-05-13
ENT-1783: Replace edx_rbac.utils.get_decoded_jwt_from_request with edx_rest_framework_extensions.auth.jwt.cookies.get_decoded_jwt

saleem-latif:
**Description:** 
This PR removes the usages of `get_decoded_jwt_from_request` from rbac in favor of `get_decoded_jwt` from edx-drf-extensions. It also adds `get_request_or_stub` and updates the usages.


enterprise_data/utils.py
```diff
@@ -57,3 +60,32 @@ def get_cache_key(**kwargs):
     key = '__'.join(['{}:{}'.format(item, value) for item, value in six.iteritems(kwargs)])
 
     return hashlib.md5(key.encode('utf-8')).hexdigest()
+
+
+def get_request_or_stub():
```
brittneyexline: instead of adding this method back in, which i think is designed to appease tests, we should call `crum.get_current_request()` directly when the request is needed.

enterprise_data_roles/rules.py
```diff
@@ -29,7 +26,7 @@ def request_user_has_implicit_access(*args, **kwargs):  # pylint: disable=unused
     __, __, request_kwargs = resolve(request.path)
```
brittneyexline, line 26: in the line above, replace with `request = crum.get_current_request()`, then when test mocking is needed, mock that function call out.

saleem-latif:
@brittneyexline PR is ready for another review.

mushtaqak (approved):
Looks good 👍

muhammad-ammar (approved):
👍

brittneyexline (approved):
can you squash your commits before merging?

PR #146, opened 2019-08-27
ENT-2245 Add enterprise data flow diagrams

crice100:
If you click "display the rich diff" in the header of the png files, you'll see the images.


crice100:
Added to the wiki: https://openedx.atlassian.net/wiki/spaces/SOL/pages/1051459601/Enterprise+data+flow

PR #147, opened 2019-09-06
ENT-1402: Replaced `has_passed` field in enrollments API with `progress_status`

ziafazal:
Replaced `has_passed` field in enrollments API with `progress_status`.


HammadAhmadWaqas (approved):
:+1:

georgebabey (approved):

PR #153, opened 2019-11-18
ENT-2387 |  Removing the NotFound exception from enrollment endpoint.

hasnain-naveed:


hasnain-naveed:
@ziafazal coverage decrease is due to removing the code `ensure_data_exists `.

ziafazal (approved):
👍

PR #159, opened 2019-12-12
ENT-1983 | Search for user in admin portal

HammadAhmadWaqas:
added search_email param in enterprise enrollment API.

### DIFF
--- a/enterprise_data/api/v0/views.py
+++ b/enterprise_data/api/v0/views.py
@@ -112,6 +112,10 @@ class EnterpriseEnrollmentsViewSet(EnterpriseViewSet, viewsets.ModelViewSet):
         elif learner_activity_param == 'inactive_past_month':
             queryset = self.filter_inactive_learners(queryset, past_month_date)
 
+        search_email = query_filters.get('search')
+        if search_email:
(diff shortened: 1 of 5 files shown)


enterprise_data/api/v0/views.py
```diff
@@ -112,6 +112,10 @@ def apply_filters(self, queryset):
         elif learner_activity_param == 'inactive_past_month':
             queryset = self.filter_inactive_learners(queryset, past_month_date)
 
+        search_email = query_filters.get('search')
+        if search_email:
+            queryset = queryset.filter(user_email__icontains=search_email)
```
ziafazal, line 117: I think this should be exact match instead of partial match

ziafazal (changes requested):
@HammadAhmadWaqas One change I suggested.

enterprise_data/api/v0/views.py
HammadAhmadWaqas: @ziafazal  As discussed in the meeting, a partial match is more feasible here.

ziafazal (approved):

georgebabey (approved):

PR #160, opened 2019-12-19
ENT-2444 | Changing coursegraph query to better capture urls

christopappas:
The updating of the query to more thoroughly match the strings we're looking for (http:// or https://) has made the resulting CSVs so large that they were unable to be emailed. As a result I have added logic to split the aggregate report into 2 parts and email both in turn, but am opting to remove the exhaustive report altogether (way too large).

### DIFF
--- a/enterprise_reporting/external_resource_link_report.py
+++ b/enterprise_reporting/external_resource_link_report.py
@@ -24,7 +24,6 @@ logging.basicConfig(level=logging.INFO)
 LOGGER = logging.getLogger(__name__)
 
 AGGREGATE_REPORT_CSV_HEADER_ROW = u'Course Key,Course Title,Partner,External Domain,Count\n'
-EXHAUSTIVE_REPORT_CSV_HEADER_ROW = u'Course Key,Course Title,Partner,External Links\n'
 
 
 def create_csv_string(processed_results, header_row, additional_columns):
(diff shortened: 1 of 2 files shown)


enterprise_reporting/external_resource_link_report.py
```diff
@@ -166,59 +174,73 @@ def query_coursegraph():
...
+
+    from_email = os.environ.get('SEND_EMAIL_FROM')
+    today = str(date.today())
     subject = 'External Resource Link Report'
     body = '''Dear Customer Success,
 Find attached a file containing course keys and their respective
 external resource links.
 
 If you have any questions/concerns with the report, please ask the
-Enterprise Team (kindly)!
```
christopappas, line 207: I don't know why I added the word kindly originally, so I'm removing it

enterprise_reporting/external_resource_link_report.py
```diff
@@ -166,59 +174,74 @@ def query_coursegraph():
     return results.data()
 
 
-def generate_and_email_report():
+def split_up_results(processed_results_part1):
     """
-    Generates a report an sends it as an email with an attachment
+    Splits up a dictionary of processed_results into a list of
+    two smaller dictionaries.
```
brittneyexline, line 180: what if we need more than two result sets at some point? seems like we should have a max size for a result set and chunk based on that, though choosing that value will probably take some guessing.

brittneyexline (commented):
just one concern around how we're splitting up the data, but otherwise looks good!

enterprise_reporting/external_resource_link_report.py
christopappas: glad you asked :D -- i actually did think about doing this but wasn't sure if i would be overengineering it (cause each email is only 4ish MB (and we can't go above 10MB, aka lots of head room)

i would do a sys.getsizeof the processed results dictionary, and based on that could determine how many parts to split it up into

brittneyexline (approved):

PR #162, opened 2019-12-31
update edx-rbac package

zamanafzal:
Update edx-rbac package and pinned django-model-utils due to the django conflicting versions.


irfanuddinahmad (approved):
LGTM!

PR #171, opened 2020-03-04
[BD] Separate reporting requirements

morenol:
### Description

Separate reporting requirements from enterprise-data requirements. Currently, enterprise-data, enterprise-data-roles and enterprise_reporting requirements are in the same file (base.in). Since enterprise_reporting is not an app that can be used by other dependencies, I think that we can separate these dependencies from base.in in order to not force edx-analytics-data-api to install these packages.

According to the README, enterprise_reporting folder contains a set of scripts used to push enterprise data reports. It supports multiple delivery methods (email, sftp) and is triggered through jenkins scheduled jobs.

I changed the tests to do not run the reporting tests if TESTENV is data.

### Constraints changed
Removed constraint in paramiko http://www.paramiko.org/changelog.html

### Reviewers
- [ ] @andrey-canon
- [x] Is this ready for edX's review?
- [ ] @jmbowman


openedx-webhooks:
Thanks for the pull request, @morenol! I've created [OSPR-4172](https://openedx.atlassian.net/browse/OSPR-4172) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

morenol:
@jmbowman This is a WIP, but I would like to know if you have any feedback about this. Should I continue with this?

jmbowman:
Sure, that makes sense as long as the requirements files output by `make upgrade` stay roughly the same (so existing scripts in other repos continue working).  It looks like unicodecsv and cryptography are needed by edx-analytics-data-api anyway (and boto3 will be once it switches from boto), but that doesn't prevent them from being removed from the core dependencies here.  Maybe add an `extras_require` in setup.py for the optional reporting dependencies, parsed from `reporting.in`?  https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

morenol:
@jmbowman I think that after this we can do the new release. All the .txt files generated with make upgrade are the same (except requirements/test-reporting.txt that has the reporting packages now).

I upgraded paramiko and pyOpenSSL but I think that  pyopenssl is not used at all.

I am not sure how you install the requirements in your jenkins scripts, if you do that using the .txt files I don't have any concerns with this changes. If that is done with pip install edx-enterprise-data, then probably you should need to change that with pip install edx-enterprise-data[reporting]

jmbowman (commented):
The Jenkins jobs install dependencies from the requirements files, so we should be good there.

I asked about pyOpenSSL, and the working theory is that it had been a dependency for a different library that was used before settling on py2neo for https://github.com/edx/edx-enterprise-data/pull/90 .  So it should be ok to remove that dependency.

Do you want to go ahead and bump the version number in this PR, or create a separate one for that?

morenol:
done @jmbowman, I removed pyopenssl and made a commit to bump the version. There is a new version of https://github.com/edx/edx-drf-extensions, so that was upgraded.

jmbowman (approved):

jmbowman:
The new release is up: https://pypi.org/project/edx-enterprise-data/

PR #178, opened 2020-03-20
ENT-2678 | Excluded the users which have only audit enrollments if 'E…

hasnain-naveed:
…nable audit data reporting' is not enabled.

### DIFF
--- a/enterprise_data/filters.py
+++ b/enterprise_data/filters.py
@@ -85,3 +85,25 @@ class AuditEnrollmentsFilterBackend(filters.BaseFilterBackend, FiltersMixin):
             queryset = queryset.exclude(**filter_query)
 
         return queryset
+
+
+class AuditUsersEnrollmentFilterBackend(filters.BaseFilterBackend, FiltersMixin):
+    """
+    Filter backend to filter Enterprise Users with 'audit' mode enrollments.
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        """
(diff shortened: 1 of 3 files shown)


enterprise_data/filters.py
```diff
@@ -85,3 +85,29 @@ def filter_queryset(self, request, queryset, view):
...
+        """
+        enterprise_id = view.kwargs['enterprise_id']
+
+        self.update_session_with_enterprise_data(request)
+
+        enable_audit_data_reporting = request.session['enable_audit_data_reporting'].get(enterprise_id, False)
+
+        enterprise_user_ids = []
+        if not enable_audit_data_reporting:
+            for record in queryset:
```
ziafazal: we should avoid this looping and use joins by filtering queryset
`queryset.filter(~Q(enrollments__user_current_enrollment_mode='audit')`

enterprise_data/models.py
```diff
@@ -126,3 +126,21 @@ def __repr__(self):
         Return uniquely identifying string representation.
         """
         return self.__str__()
+
+    @property
+    def has_paid_mode_enrollment(self):
```
ziafazal: This is not needed if make above queryset change.

ziafazal (changes requested):
@hasnain-naveed completed first round of review.

ziafazal (approved):
👍

moconnell1453 (approved):

PR #182, opened 2020-03-23
[BD] Add tests using django2

morenol:
## Description

Add tests using django 2.0, 2.1 and 2.2. In the case of the reporting scripts only were added the django 2.2 tests since I think that is is not necessary to test the other versions. 

Part of https://openedx.atlassian.net/projects/BOM/issues/BOM-1359.

## Reviewers

- [ ] @andrey-canon
- [x] Ready for edX review
- [x] @jmbowman

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ env:
   - TOXENV=django111-data
   - TOXENV=django111-reporting
   - TOXENV=quality
+  - TOXENV=django20-data
+  - TOXENV=django21-data
+  - TOXENV=django22-data
+  - TOXENV=django22-reporting
 cache:
   - pip
 install:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,10 +13,14 @@ Change Log
 
 Unreleased
 ----------
-* Fix for JWT being double encoded
-* Drop python 2.7 support
 
 =========================
+
+[2.0.0] - 2020-04-01
+--------------------
+* Fix for JWT being double encoded
+* Drop python 2.7 support
+* Add support to Django 2.0, 2.1 and 2.2
 
 [1.3.16] - 2020-03-13
 ---------------------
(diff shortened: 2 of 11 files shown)


openedx-webhooks:
Thanks for the pull request, @morenol! I've created [OSPR-4294](https://openedx.atlassian.net/browse/OSPR-4294) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

jmbowman (approved):
@macdiesel Note that as written this would have the two enterprise reports in https://github.com/edx/jenkins-job-dsl-internal/blob/master/jobs/tools-edx-jenkins.edx.org/createEnterpriseDataReportingJobs.groovy start to run with Django 2.2 after merging, but edx-analytics-data-api could keep using Django 1.11 until the rest of the code is updated.  Let me know if you anticipate any problems with that.

macdiesel:
I'm a bit confused, this PR would mean we have to upgrade insights to django 2.2?

morenol:
@macdiesel no, insights and data-api could keep using Django 1.11(Django 1.11 is still on the tests), the difference now is that all the scripts that are in https://github.com/edx/edx-enterprise-data/tree/master/enterprise_reporting  will start  to run Django 2.2 because `requirements/base.txt` now has Django==2.2

morenol:
If any of you think that we should keep with Django==1.11 as the main version of django for this library, let me know and I change this PR, I want to get this merged in order to create a new release to solve the problem with enterprise-data in data api.

morenol:
I made Django==1.11 as the default django, so this might get merged faster.

macdiesel:
I'm fine with going with Django 2 if those PRs are ready we can deploy this to the prod systems. Just trying to understand.

jmbowman:
@macdiesel This repo is used both as a dependency of edx-analytics-data-api and as the home of code used to run enterprise reporting scripts in Jenkins.  The suggestion was that we could upgrade the latter to Django 2.2 now without much consequence, while making sure it still works as a service dependency using either Django version.  But we decided to just stick with 1.11 for now until everything's ready to upgrade to 2.2.

jmbowman:
https://pypi.org/project/edx-enterprise-data/2.0.0/

PR #191, opened 2020-03-27
[BD] Set enterprise_user_id as unique

morenol:
## Description

Migrations in mysql(for reports database) are failing due to enterprise_user_id is not set as unique (in the [model](https://github.com/edx/edx-enterprise-data/blob/88f808b6bbdd047e36e865f01aae2a53df501bc7/enterprise_data/models.py#L106) it is defined at that way). More context in https://github.com/edx/edx-analytics-data-api/pull/340#discussion_r399545972

## Reviewers
- [ ] @andrey-canon
- [x] Ready for edX review.
- [ ] @jmbowman

### DIFF
--- a/enterprise_data/migrations/0011_enterpriseuser.py
+++ b/enterprise_data/migrations/0011_enterpriseuser.py
@@ -18,7 +18,7 @@ class Migration(migrations.Migration):
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('enterprise_id', models.UUIDField()),
                 ('lms_user_id', models.PositiveIntegerField()),
-                ('enterprise_user_id', models.PositiveIntegerField()),
+                ('enterprise_user_id', models.PositiveIntegerField(unique=True)),
                 ('enterprise_sso_uid', models.CharField(max_length=255, null=True)),
                 ('user_account_creation_timestamp', models.DateTimeField(null=True)),
                 ('user_email', models.CharField(max_length=255, null=True)),


openedx-webhooks:
Thanks for the pull request, @morenol! I've created [OSPR-4323](https://openedx.atlassian.net/browse/OSPR-4323) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

morenol:
It seems that builds are frozen https://travis-ci.org/github/edx/edx-enterprise-data/pull_requests

jmbowman:
@macdiesel Is there an existing production database for this that would be unhappy about adding this constraint to an existing field?  If so, how did it get around this migration issue?  Is the analytics database actually always SQLite like in devstack?

morenol:
@jmbowman In my experience, the times that I have installed this service, I have installed this using mysql for that database. I think that in prod we don't have that problem because we don't run `python manage.py migrate --database=analytics`, instead these tables are created by edx-analytics-pipeline:

https://github.com/edx/edx-analytics-pipeline/blob/01c4980f0a5026873a23b556448ee7499b0c8cb9/edx/analytics/tasks/enterprise/enterprise_user.py#L33

https://github.com/edx/edx-analytics-pipeline/blob/90c947087ffb34521f684acc021075fe2581cc19/edx/analytics/tasks/warehouse/load_warehouse_insights.py#L177

brianhw:
@morenol I'm oncall, and picking up this review now.   My experience is more on the pipeline side.  The load_warehouse_insights code shouldn't be a direct factor -- that's a copy we currently make to our Vertica warehouse, and is not "production" in the same sense.  

So the task where the table is loaded is actually https://github.com/edx/edx-analytics-pipeline/blob/01c4980f0a5026873a23b556448ee7499b0c8cb9/edx/analytics/tasks/enterprise/enterprise_user.py#L161-L194.   There is currently no hook in this code to specify uniqueness constraints.   We could add that, and add it to the definition of this table, for the next time someone creates the table (e.g. open source community, perhaps).   But we do not have any concept of migrations for the tables in this database -- the pipeline will create the table if it doesn't exist, but otherwise just writes to it.  So if we want a "migration" to go against our production server, we have to apply it manually.    Is that what needs to be done before this commit is merged?   

And also, just so I'm clear, the migration here is for keeping a development version of the table in sync with what is produced by the pipeline, so that fake data can be loaded into it for development and testing.  And this had not been a problem in the past because such development had been using SQLite, which didn't complain about the missing uniqueness constraint.   Is that right?  So this migration PR is blocking other work to switch the DB from SQLite to be mysql for everything?

morenol:
@brianhw Yes, we only need this to get insights and data-api working on the devstack and therefore using mysql. 

I think that we don't need any changes to the production environment since that migration files are only used in order to generate fake data for the development environment. 

https://github.com/edx/edx-analytics-data-api/blob/d0beecc7d527133aec6aa493a25482b112a7fcfb/Makefile#L107

The other option is to ignore this and connect directly in the devstack the pipeline with insights but I think that in any case this change wouldn't affect negatively anything and would allow the possibility to run insights and data api in the devstack without the need of more containers and services running.

brianhw:
@morenol We've got a separate devstack (called analytics-devstack) that spins up a Hadoop cluster to allow the pipeline to be run.  I don't know of plans to fold that into devstack -- it takes more resources, so it would need to be optional at any rate.   So I think the connecting this to the pipeline is not a short-term option, and probably not a great option at all, given the difficulty in setting it up and feeding it data and all.

Oh, I was actually more worried about a mismatch between the Django model and the underlying production MySQL db that it would be pointing at.   But the model has had this uniqueness key in place for over two years, so it should not be an issue now if it wasn't before. 

So, LGTM!

brianhw (approved):
LGTM.   👍

PR #198, opened 2020-05-20
requirements target to use piptools as source to ensure correct setup

binodpant:
WHY: This fixes the issue I saw when I first ran `make requirements` on this repo, without running the piptools command which was not mentioned in the doc

WHAT: Just added piptools target and used it as source for requirements target


Makefile
```diff
@@ -14,6 +14,9 @@ clean: ## remove generated byte code, coverage reports, and build artifacts
 	rm -fr dist/
 	rm -fr *.egg-info
 
+piptools: ## install tools prior to requirements
```
bbaker6225: Nit: I'd call this something like `piptools-requirements` or `requirements.piptools`, etc.

bbaker6225: Although I see that this is the pattern other repos use, so sure seems fine

bbaker6225 (approved):
Hm not sure what's going on with that failing test, but LGTM once that is fixed

binodpant:
Thanks not sure what that is, looking, seems unrelated to the makefile change

Makefile
```diff
@@ -40,7 +43,7 @@ upgrade: ## update the requirements/*.txt files with the latest packages satisfy
 	mv requirements/test-reporting.tmp requirements/test-reporting.txt
 
 
-requirements: ## install development environment requirements
+requirements: piptools-requirements ## install development environment requirements
```
kellmarie, line 46: [NIT] I would move `piptools-requirements` down a line and leave `requirements: ## install development environment requirements` as is

binodpant: Ah thanks for the review, I think there is a difference though. If I place it here it is a source for the target, whereas if I place it down a line, will it still be source or part of commands? I have to check that but I thought making it source is better to express a dependency from requirements to piptools-requirements, what do you think?

binodpant:
Just updating, a rerun made this pass that build so certainly it's a flaky issue. I am still going through and trying to understand the test setup to try to perform cleanup as per @iloveagent57 's recommendation, but I will try and pick it up once I am through with the training cycles. If you prefer I merge this now and look at the test issue separately with a new ticket I can do so too.

binodpant:
I will merge this in for cleanliness. I think it is better to work on the flaky test issue separately which we can track with a ticket potentially

PR #205, opened 2020-07-20
Update EnterpriseEnrollmentFactory

raq929:
Adds data needed to have full data on the dashboard.

The only test I'm a little worried about changing is on line 993. I couldn't quite understand what the result was supposed to be based on the filters. I'm fairly certain that my changes should update the count there, but not 100% sure. Advice on how to be sure this test is testing the right thing would be helpful.

### DIFF
--- a/enterprise_data/tests/test_views.py
+++ b/enterprise_data/tests/test_views.py
@@ -360,6 +360,7 @@ class TestEnterpriseEnrollmentsViewSet(JWTTestMixin, APITransactionTestCase):
     )
     @ddt.unpack
     def test_get_queryset_returns_enrollments_with_learner_activity_filter(self, activity_filter, expected_dates):
+
         enterprise = EnterpriseUserFactory()
         enterprise_id = enterprise.enterprise_id
         url = "{url}?learner_activity={activity_filter}".format(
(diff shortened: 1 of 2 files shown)


enterprise_data/tests/test_views.py
```diff
@@ -953,14 +988,15 @@ def test_viewset_filter_active_courses_true(self):
             'v0:enterprise-users-list',
             kwargs=kwargs,
         )
+
         response = self.client.get(url, params)
-        assert response.json()['count'] == 2
+        assert response.json()['count'] == 4
```
muhammad-ammar, line 993: @raq929 This test is verifying [this filtering](https://github.com/edx/edx-enterprise-data/blob/98e1000d8aa10afa23015ad8605e5dd1fb355bd9/enterprise_data/api/v0/views.py#L246-L251) in the view. For `'active_courses': 'true'`, api will return learners enrolled in courses having end date in future.

muhammad-ammar (approved):
LGTM 👍 

one question though: Can you help me understand why we are updating the factory?

raq929:
Yeah, I needed a lot of dummy data for my manual testing, and having a more fully functional factory seemed like an easy way to get there quickly. I may follow up with a management command that creates a lot of users and enrollments.

PR #206, opened 2020-07-22
Add course title and start date filtering

raq929:
Backend support for ENT-3086.
Adds filtering and tests.

### DIFF
--- a/enterprise_data/tests/test_views.py
+++ b/enterprise_data/tests/test_views.py
@@ -213,6 +213,44 @@ class TestEnterpriseEnrollmentsViewSet(JWTTestMixin, APITransactionTestCase):
         assert response['count'] == expected_count
         assert response == expected_response
 
+    @ddt.data(
+        ('keep', 2),
+        ('Beekeep', 1),
+        ('101', 2)
+    )
+    @ddt.unpack
+    def test_get_enterprise_enrollments_with_course_search(self, search_course, expected_count):
+        enterprise_user = EnterpriseUserFactory(enterprise_id=self.enterprise_id)
+        course_titles = ["Beekeeping 101", "Keeping it together", "Bears and their mountains 101"]
+        for course_title in course_titles:
+            EnterpriseEnrollmentFactory(
+                enterprise_user=enterprise_user,
+                enterprise_id=self.enterprise_id,
+                course_title=course_title,
+                has_passed=False,
+            )
+        enrollments = EnterpriseEnrollment.objects.filter(
+            enterprise_id=self.enterprise_id,
+            course_title__icontains=search_course
+        )
+        enrollments = EnterpriseEnrollment.objects.filter(
+            enterprise_id=self.enterprise_id,
+            course_title__icontains=search_course)
+        expected_course_titles = [en.course_title for en in enrollments]
+        response = self.client.get(
+            reverse('v0:enterprise-enrollments-list', kwargs={'enterprise_id': self.enterprise_id}),
+            data={'search_course': search_course}
+        ).json()
+        response = self.client.get(
+            reverse('v0:enterprise-enrollments-list', kwargs={'enterprise_id': self.enterprise_id}),
+            data={'search_course': search_course}
+        ).json()
+
+        assert response['count'] == expected_count
+
+        for result in response["results"]:
+            assert result["course_title"] in expected_course_titles
+
     def test_get_queryset_returns_no_enrollments(self):
         """ Test that enterprise with no enrollments returns empty list """
         enterprise = EnterpriseUserFactory()
--- a/enterprise_data/tests/test_views.py
+++ b/enterprise_data/tests/test_views.py
@@ -3,7 +3,6 @@
 Tests for views in the `enterprise_data` module.
 """
 
-
 from datetime import date, datetime, timedelta
 
 import ddt
@@ -18,7 +17,7 @@ from django.utils import timezone
 
 from enterprise_data.api.v0.serializers import EnterpriseEnrollmentSerializer
 from enterprise_data.api.v0.views import subtract_one_month
-from enterprise_data.models import EnterpriseEnrollment
+from enterprise_data.models import EnterpriseEnrollment, EnterpriseUser
 from enterprise_data.tests.mixins import JWTTestMixin
 from enterprise_data.tests.test_utils import (
     EnterpriseEnrollmentFactory,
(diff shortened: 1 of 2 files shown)


enterprise_data/tests/test_views.py
```diff
@@ -213,6 +214,185 @@ def test_get_enterprise_enrollments_with_email_search(self, search_email, expect
...
+
+        EnterpriseEnrollmentFactory(
+                enterprise_user=enterprise_user1,
+                enterprise_id=self.enterprise_id,
+                course_start=course_start_dates[0],
+            )
+        EnterpriseEnrollmentFactory(
+                enterprise_user=enterprise_user2,
+                enterprise_id=self.enterprise_id,
+                course_start=course_start_dates[0],
```
taliaedX: I'm confused. Why are you looping through this `course_start_dates` list making an EnterpriseEnrollmentFactory object and then manually appending each element from the course_start_date list outside of this forloop? I'm guessing because different enterprise user objects, but it seems like two different implementations are being used here (loop vs manual index fetching in the list)

Did you intend to do something like (I put enterprise_users in a list because a nested loop here just seems appropriate for a change):

```
enterprise_users = [enterprise_user0, enterprise_user1, enterprise_user2]
for start_date in course_start_dates:
    for user in enterprise_users:
        EnterpriseEnrollmentFactory(
            enterprise_user=user,
            enterprise_id=self.enterprise_id,
            course_start=start_date,
        )
```

enterprise_data/tests/test_views.py
```diff
@@ -213,6 +214,185 @@ def test_get_enterprise_enrollments_with_email_search(self, search_email, expect
...
+
+        response = self.client.get(
+            reverse('v0:enterprise-enrollments-list', kwargs={'enterprise_id': self.enterprise_id}),
+            data={'search_start_date': search_date}
+        ).json()
+
+        assert response['count'] == expected_count
+
+        for result in response["results"]:
+            assert pytz.utc.localize(datetime.strptime(result["course_start"], "%Y-%m-%dT%H:%M:%SZ")) \
```
taliaedX: what is the purpose of localization here?

enterprise_data/tests/test_views.py
```diff
@@ -213,6 +214,185 @@ def test_get_enterprise_enrollments_with_email_search(self, search_email, expect
...
+        ('2020-09-03T00:00:00Z', 2),
+        ('2020-08-30T00:00:00Z', 1)
+    )
+    @ddt.unpack
+    def test_get_enterprise_enrollments_with_start_date_search(self, search_date, expected_count):
+        enterprise_user0 = EnterpriseUserFactory(enterprise_id=self.enterprise_id)
+        enterprise_user1 = EnterpriseUserFactory(enterprise_id=self.enterprise_id)
+        enterprise_user2 = EnterpriseUserFactory(enterprise_id=self.enterprise_id)
+        course_start_dates = [
+            pytz.utc.localize(datetime(2020, 9, 30)),
```
taliaedX: I went backwards with this, but why localize here. Are you just trying to get a datetime that actually includes timezone? Or the opposite?

enterprise_data/tests/test_views.py
```diff
@@ -213,6 +214,185 @@ def test_get_enterprise_enrollments_with_email_search(self, search_email, expect
...
+        for title, email, start_date in zip(course_titles, emails, course_start_dates):
+            EnterpriseEnrollmentFactory(
+                enterprise_user=enterprise_user0,
+                enterprise_id=self.enterprise_id,
+                course_title=title,
+                course_start=start_date,
+                user_email=email
+            )
+
+        EnterpriseEnrollmentFactory(
```
taliaedX: again I feel like this wasn't supposed to be outside this loop but I may be taking crazy pills.

enterprise_data/tests/test_views.py
```diff
@@ -213,6 +214,185 @@ def test_get_enterprise_enrollments_with_email_search(self, search_email, expect
...
+            "enterprise_id": self.enterprise_id
+        }
+        if search_email:
+            search_kwargs["user_email__icontains"] = search_email
+        if search_course:
+            search_kwargs["course_title__icontains"] = search_course
+        if search_date:
+            search_kwargs["course_start"] = search_date
+
+        enrollments = EnterpriseEnrollment.objects.filter(**search_kwargs)
```
taliaedX: The whole structure of this test feels odd to me, but not sure how to pin it down to one specific thing. The fact you are basically having to copy the logic out the `apply_filters` function feels very weird to me. There's a lot of looping here as well and I wonder if perhaps either the ddt data parameters should be a bit more extensive or if we can consolidate data down to some static vars.

taliaedX (changes requested):
So the actual query part in the views is a-okay.

The tests are where I might be losing my mind.
The structure seems a bit odd and there's some spots where it may be obvious (or I may be losing my mind) what is off and others where I'm not entirely sure, but it might help if we can chat over what the intended test cases are and what data is absolutely necessary for them to make sure I can either understand the existing test structure or help figure out something maybe more efficient.

enterprise_data/tests/test_views.py
raq929: The purpose of localization is not getting an warning every time I make an enrollment. 😉

enterprise_data/tests/test_views.py
raq929: No, it is. _Sigh_. I'm not in love with this.

enterprise_data/tests/test_views.py
raq929: Yeah. Let's chat tomorrow morning and see what we come up with. 
I don't love the copying the logic either; it's something I borrowed from other tests in this test suite, but I agree it does not give me a good feeling.

enterprise_data/tests/test_views.py
raq929: I wanted to be sure I have enrollments for each of my course titles/ start dates/ emails, and one thing that has some of those things again. To be sure that it works with more than one enrollment.
You're not losing your mind, these tests are kind of based on other tests in this test suite, but have requirements that went beyond the ability of that structure to handle it.

enterprise_data/tests/test_views.py
taliaedX: Now I'm just curious what warning was happening lol

enterprise_data/tests/test_views.py
```diff
@@ -1666,3 +1650,194 @@ def test_permissions_with_enterprise_openedx_operator(self):
...
+        expected_titles = [title for title in self.all_course_titles if search_course.lower() in title.lower()]
+        response = self.client.get(
+            reverse('v0:enterprise-enrollments-list', kwargs={'enterprise_id': self.enterprise_id}),
+            data={'search_course': search_course}
+        ).json()
+        response = self.client.get(
+            reverse('v0:enterprise-enrollments-list', kwargs={'enterprise_id': self.enterprise_id}),
+            data={'search_course': search_course}
+        ).json()
+        # import pdb; pdb.set_trace()
```
taliaedX: I know you saw it on the call, but just to doubly make sure that this gets removed. :p

taliaedX (approved):
So per our call, this is definitely a bit neater and makes the test cases a bit clearer. We could still use a little info on why the total expected enrollment counts are as such whether that be in each individual test or commented with the data. But it's definitely a lot nicer and as before is quite thorough, so definitely mergeable!

enterprise_data/tests/test_views.py
raq929: A very large annoying one about how I gave Django a native datetime when I ought to have used a timezone blah blah blah

PR #215, opened 2020-12-09
ENT-3793, ENT-3795, ENT-3797: Updated completion, course structure and grade reports for file download from S3

irfanuddinahmad:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise_reporting/clients/s3.py
+++ b/enterprise_reporting/clients/s3.py
@@ -1,0 +1,24 @@
+"""
+Client for connecting to AWS S3.
+"""
+
+import boto3
+
+
+class S3Client:
+    """
+    Client for connecting to AWS S3.
+    """
+
+    def __init__(self):
+        pass
+
+    def get_enterprise_report(self, report_name, file_handler):
+        """
+        Download a report file from S3
+        report_name <string>: name of the report on S3
+        file_handler <file>: file object opened in binary mode
+        """
+        bucket_name = 'edx-enterprise-reporting'
+        s3 = boto3.resource('s3')
+        s3.download_fileobj(bucket_name, report_name, file_handler)
--- a/enterprise_reporting/reporter.py
+++ b/enterprise_reporting/reporter.py
@@ -17,7 +17,7 @@ from enterprise_reporting.clients.enterprise import (
     EnterpriseAPIClient,
     EnterpriseDataApiClient
 )
-from enterprise_reporting.clients.snowflake import SnowflakeClient
+from enterprise_reporting.clients.s3 import S3Client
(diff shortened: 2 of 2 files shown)


enterprise_reporting/clients/s3.py
```diff
@@ -0,0 +1,22 @@
...
+
+class S3Client:
+    """
+    Client for connecting to AWS S3.
+    """
+
+    def __init__(self):
+        pass
+
+    def get_file(self, S3_object_name, file_handler):
```
saleem-latif: nit: I think we should use snake case here i.e. rename to `s3_object_name `

enterprise_reporting/clients/s3.py
```diff
@@ -0,0 +1,22 @@
...
+        """
+        Download a file from S3
```
saleem-latif: nit, add argument details in the doc string.

```
Arguments:
    s3_object_name (<type>): <description>
    file_handler (<type>): <description>
```

enterprise_reporting/reporter.py
```diff
@@ -394,18 +394,10 @@ def _generate_enterprise_report_course_structure_csv(self):
...
-            query = self.SNOWFLAKE_COMPLETION_QUERY.format(
-                fields=','.join(self.SNOWFLAKE_COMPLETION_QUERY_FIELDS),
-                enterprise_id=self.enterprise_customer_uuid.replace('-', '')
-            )
-            LOGGER.debug('Executing this Snowflake query: {}'.format(query))
-            data_report_file_writer.writerows(snowflake_client.stream_results(query))
-        snowflake_client.close_connection()
+        s3_client = S3Client()
+        with open(self.data_report_file_name, 'wb') as data_report_file:
+            LOGGER.debug('Fetching enterprise completion report from S3)
```
saleem-latif: would it be helpful to add more information in the debug message? like bucket name and target file name?

saleem-latif (approved):
A couple of comments otherwise looks good 👍

enterprise_reporting/clients/s3.py
irfanuddinahmad: Done

enterprise_reporting/clients/s3.py
irfanuddinahmad: Done

enterprise_reporting/reporter.py
irfanuddinahmad: 'completion' is the name of the file. We don't have bucket info here.

muhammad-ammar (approved):

PR #218, opened 2021-01-02
Revert "Python Code Cleanup"

georgebabey:
Reverts edx/edx-enterprise-data#217

bcrypt 3.2 drops python 3.5 support so our Jenkins job will need to get upgraded to python 3.8 first.


UsamaSadiq:
PR https://github.com/edx/edx-enterprise-data/pull/219 made for pinning the required packages for the job to pass and the job will, later on, be upgraded to use python38 so this revert is not needed.

PR #219, opened 2021-01-04
WIP: Pin bcrypt version <3.2

UsamaSadiq:
After merging the https://github.com/edx/edx-enterprise-data/pull/217, the [Enterprise reporting job](https://tools-edx-jenkins.edx.org/job/enterprise/job/enterprise-data-reporting/) started failing because of python version conflict. 
So, pinned the versions required by `python35` unless the job is upgraded to run with `python38` instead.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -53,8 +53,9 @@ chardet==3.0.4
     # via requests
 colorama==0.4.3
     # via awscli
-cryptography==3.3.1
+cryptography==3.2.1
     # via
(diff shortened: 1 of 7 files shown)


georgebabey (approved):
Thank you for getting these pinned until we can upgrade!

UsamaSadiq:
Upgraded the reporting jobs to run using `python38` in PR https://github.com/edx/jenkins-job-dsl-internal/pull/482 so this change is not needed now.

PR #232, opened 2021-05-19
feat: transition to snowflake powered analytics

muhammad-ammar:
**Merge checklist:**

**Post merge:**


enterprise_data/api/v1/views.py
```diff
@@ -0,0 +1,357 @@
...
+    Viewset for routes related to Enterprise course enrollments.
+    """
+    serializer_class = serializers.EnterpriseEnrollmentSerializer
+    pagination_class = EnterpriseEnrollmentsPagination
+    filter_backends = (AuditEnrollmentsFilterBackend, ConsentGrantedFilterBackend, filters.OrderingFilter,)
+    ordering_fields = '__all__'
+    ordering = ('user_email',)
+    # Remove this filter. We will show all the results. For learners who
+    # has not given DSC, their personal information will be anoymized.
+    CONSENT_GRANTED_FILTER = 'is_consent_granted'
```
muhammad-ammar: This filter needs to be removed to send data to enterprise always but in case if use has not given consent, data will be anonymized.

enterprise_data/api/v1/serializers.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+
+    class Meta:
+        model = EnterpriseLearner
+        exclude = ('created_at', )
+
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+
+        if hasattr(instance, 'enrollment_count'):
```
saleem-latif, line 57: why do we need this?

enterprise_data/api/v1/views.py
```diff
@@ -0,0 +1,357 @@
...
+    AuditUsersEnrollmentFilterBackend,
+    ConsentGrantedFilterBackend,
+)
+from enterprise_data.models import EnterpriseLearner, EnterpriseLearnerProgress
+from enterprise_data.paginators import EnterpriseEnrollmentsPagination
+
+LOGGER = getLogger(__name__)
+
+
+def subtract_one_month(original_date):
```
saleem-latif, line 33: nit: I think you can simplify this
```
def subtract_one_month(original_date):
    """
    Returns a date exactly one month prior to the passed in date.
    """
    if original_date.month > 1:
        return original_date.replace(month=original_date.month-1)
    else:
        return original_date.replace(year=original_date.year - 1, month=12)
```

enterprise_data/api/v1/serializers.py
muhammad-ammar: `enrollment_count` is set by query annotation and will be available for specific filters that is why we are adding it conditionally.

enterprise_data/api/v1/views.py
muhammad-ammar: * This is a copy/paste of https://github.com/edx/edx-enterprise-data/blob/09cc8b49c08f0cf4247b2a9ad585adb7bb0ac0be/enterprise_data/api/v0/views.py#L34-L42
* Why we need to handle the first month separately?
* I think a more simple solution is https://stackoverflow.com/a/9725091/216514 copy/pasting the answer below
```python
>>> import datetime 
>>> import dateutil.relativedelta
>>> now = datetime.datetime.now()
>>> print now
2012-03-15 12:33:04.281248
>>> print now + dateutil.relativedelta.relativedelta(months=-1) 
2012-02-15 12:33:04.281248
```

saleem-latif (approved):

PR #259, opened 2021-08-13
[Fix] Update the ref_name of the serializers with the same name in v0 and v1 of the enterprise data api

schenedx:
The reason this needs to be done is because, drf_yasg, which is the library to generate swagger docs, have trouble with the same name of the two serializers in v0 and v1 of the API version

**Merge checklist:**

**Post merge:**
- [x] Tag pushed and a new [version](https://github.com/edx/edx-enterprise-data/releases) released
- [x] After versioned build finishes in [Travis](https://travis-ci.org/github/edx/edx-enterprise-data), verify version has been pushed to [PyPI](https://pypi.org/project/edx-enterprise-data/)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,10 @@ Unreleased
 ----------
 
 =========================
+[2.2.20] - 2021-08-13
+---------------------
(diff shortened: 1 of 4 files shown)


muhammad-ammar (approved):
LGTM 👍

PR #275, opened 2021-10-21
feat!: advertise constraints when packaging

rgraber:
Updates setup.py to read constraints files as well when generating metadata about requirements. This will enable library users to more quickly see if an upgrade is compatible with current requirements.

Before, edx_enterprise_data.egg-info/requires.txt looked like:
```
edx-django-utils
requests
rules
edx-drf-extensions
edx-rbac
django-fernet-fields
edx-opaque-keys
edx-rest-api-client
Django
django-model-utils

[reporting]
paramiko
PGPy
celery==3.1.18
vertica-python
pyminizip
snowflake-connector-python
py2neo
awscli==1.18.159
boto3==1.15.18
unicodecsv==0.14.1
cryptography
```

After the change:
```
requests
edx-rest-api-client
edx-django-utils
edx-drf-extensions
edx-opaque-keys
Django<2.3,>=2.2
django-fernet-fields
django-model-utils
edx-rbac
rules

[reporting]
awscli==1.18.159
boto3==1.15.18
celery==3.1.18
cryptography
paramiko
PGPy
py2neo
pyminizip
unicodecsv==0.14.1
vertica-python
snowflake-connector-python
```

Note that the constraints on Django are now listed

**Merge checklist:**
- [N/A ] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files)
- [N/A ] `make static` has been run to update webpack bundling if any static content was updated
- [ N/A] Translations updated (see docs/internationalization.rst but also this isn't blocking for merge atm)

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,10 @@ Unreleased
 ----------
 
 =========================
+[3.4.0] - 2021-10-22
+---------------------
+  * Advertise constraints when packaging
+
 [3.3.0] - 2021-09-21
 ---------------------
   * Added support for Django32
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,8 @@ import sys
 
 from setuptools import setup
 
+# groups "my-package-name<=x.y.z,..." into ("my-package-name", "<=x.y.z,...")
+requirement_line_regex = re.compile(r"([a-zA-Z0-9-]+)([<>=][^#\s]+)?")
 
 def get_version(*file_paths):
     """
@@ -31,17 +33,46 @@ if sys.argv[-1] == "tag":
 
 def load_requirements(*requirements_paths):
     """
-    Load all requirements from the specified requirements files.
+    Load all requirements from the specified requirements files, including any constraints from other files that
+    are pulled in.
     Returns a list of requirement strings.
     """
-    requirements = set()
+    requirements = {}
+    constraint_files = set()
+
+    def add_version_constraint_or_raise(current_line, current_requirements, add_if_not_present):
+        regex_match = requirement_line_regex.match(current_line)
+        if regex_match:
+            package = regex_match.group(1)
+            version_constraints = regex_match.group(2)
+            existing_version_constraints = current_requirements.get(package, None)
+            # it's fine to add constraints to an unconstrained package, but raise an error if there are already
+            # constraints in place
+            if existing_version_constraints and existing_version_constraints != version_constraints:
+                raise BaseException('Multiple constraint definitions found for {}: "{}" and "{}".'
+                                    .format(package, existing_version_constraints, version_constraints))
+            if add_if_not_present or package in current_requirements:
+                current_requirements[package] = version_constraints
+
+    # process .in files and store the path to any constraint files that are pulled in
     for path in requirements_paths:
         with open(path) as reqs:
-            requirements.update(
-                line.split('#')[0].strip() for line in reqs
-                if is_requirement(line.strip())
-            )
-    return list(requirements)
+            for line in reqs:
+                if is_requirement(line):
+                    add_version_constraint_or_raise(line, requirements, True)
+                # ignore non-local constraint files
+                if line and line.startswith('-c') and not line.startswith('-c http'):
+                    constraint_files.add(os.path.dirname(path) + '/' + line.split('#')[0].replace('-c', '').strip())
+
+    # process constraint files and add any new constraints found to existing requirements
+    for constraint_file in constraint_files:
+        with open(constraint_file) as reader:
+            for line in reader:
+                if is_requirement(line):
+                    add_version_constraint_or_raise(line, requirements, False)
+
+    # process back into list of "pkg><=constraints" strings
+    return ['{}{}'.format(pkg, version or '') for (pkg, version) in requirements.items()]
 
 
 def is_requirement(line):
@@ -49,7 +80,7 @@ def is_requirement(line):
     Return True if the requirement line is a package requirement;
     that is, it is not blank, a comment, a URL, or an included file.
     """
-    return line and not line.startswith(('-r', '#', '-e', 'git+', '-c'))
+    return line and line.strip() and not line.startswith(('-r', '#', '-e', 'git+', '-c'))
 
 
 setup(
--- a/enterprise_data/__init__.py
+++ b/enterprise_data/__init__.py
@@ -2,6 +2,6 @@
 Enterprise data api application. This Django app exposes API endpoints used by enterprises.
 """
 
-__version__ = "3.3.0"
+__version__ = "3.4.0"
 
 default_app_config = "enterprise_data.apps.EnterpriseDataAppConfig"  # pylint: disable=invalid-name


setup.py
```diff
@@ -8,6 +8,8 @@
 
 from setuptools import setup
 
+# groups "my-package-name<=x.y.z,..." into ("my-package-name", "<=x.y.z,...")
+requirement_line_regex = re.compile(r"([a-zA-Z0-9-]+)([<>=][^#\s]+)?")
```
timmc-edx, line 12: Might need to put `\s*(#.*)?` at the end for lines like this:

```requests                                # Required for SAPSuccessFactorsAPIClient```

Or maybe I'm missing something, since I haven't actually tried this locally.

Comment (timmc-edx) [reply setup.py:12]: ...I think I'm just missing something, because I've done some experiments and I don't understand what I'm seeing. -.-

Comment (rgraber) [reply setup.py:12]: since we're only using this to match the beginning of the line (eg match instead of matchfull) it should successfully catch lines like that and just have None for the second group

Comment (timmc-edx) [reply setup.py:12]: Oh jeez, yeah, `fullmatch` is how it's spelled in Python, as opposed to `match` in some other languages' stdlibs.

File: CHANGELOG.rst [hunk 2]
```diff
@@ -15,6 +15,10 @@ Unreleased
 ----------
 
 =========================
+[3.4.0] - 2021-10-22
+---------------------
+  * Advertise constraints when packaging
```
robrap, line 20: In general, I think these PRs should be considered a bug fix, since in theory they should be a no-op of sorts. In other words, this should be `3.3.1`.

However, this PR indicates a rollout gotcha of your new code. If the library starts advertising invalid constraints, other IDAs or libraries may no longer be able to update this library when it should be able to.

As noted in a separate conversation (and the PR description), this PR would start advertising `Django<2.3`, but the constraint should be `Django<4`(?, see below) to match reality. Note that [edx-analytics-data-api](https://github.com/edx/edx-analytics-data-api/blob/master/requirements/base.txt) is on `Django 3.2`, and is using `edx-enterprise-data 3.2.0`, but supposedly `Django 3.2` support was added in `edx-enterprise-data 3.3.0`, which is sad/funny.

In any case, I would propose that as part of rollout of this code in any new repo, if the advertised constraints change, they should be verified/fixed as part of the same PR. This library should be advertising `Django 3.2` support, and that could and should (in my opinion) be fixed in this same PR.

Note: In general, I think we don't create upper constraints for all dependencies for major versions because of the overhead, so I think we just add it when we know/learn that there is an issue. For large libraries like Django, maybe that is different, and we assume guilty until proven innocent? Thus, bumping to `Django<4` would make sense.

rgraber: Small point: the reason I made it a version change was specifically because it would break anyone trying to upgrade until the correct constraints were advertised. I felt that was enough of a breaking change.

rgraber: I'm a little hesitant to group this PR with the PR to correct the constraints from a rollback/debug point of view. Those feel like separate problems and if for some reason, this turns out not actually to be totally 3.2 compliant, rolling back that change should be a different PR than rolling back the change to advertising that.

rgraber: In this particular case that's unlikely, but the idea holds.

timmc-edx: The actual compatibility of the code doesn't change, which makes me think minor or patch version change at most -- but it's possible it could cause dependency version downgrades in the relying IDA. Making a major version bump would get people's attention. :-)

When we automate (or semi-automate) PR-creation for these changes, we can have the job include a before/after diff of the requires.txt output. If there's a tool to auto-bump a version number we could always do a major version bump, and anyone reviewing the PR could decide whether to remove that part of the changeset.

timmc-edx: (I don't think *we* should group these changes, but I'd like to see PRs generated and then handed off to owning teams any time the advertised constraints would change so that they can make a decision and make any adjustments themselves.)

rgraber: Put like that, I'm ok with making it a minor version change, and I like the idea of always diffing the requires.txt files. I don't understand what you mean by "I'd like to see PRs generated and then handed off to owning teams any time the advertised constraints would change." Are you talking about auto generating PRs to update constraints.txt?

timmc-edx: Sorry, what I mean is that I'd like to generate PRs for all relevant repos, and have the PRs somehow include the requires.txt diff (might need a small change to the tool) -- and that for any of those PRs where the diff in fact shows the addition of contraints, we get the owning team involved to have them approve the PR. (Maybe they want to make fixes, maybe they just say it looks good.)

rgraber: Ah, that makes sense to me.

rgraber: I guess in this case since we know what the fix would be we could just make it

rgraber: In terms of the actual version of Django to put, edx-analytics-data-api just uses Django with no version constraint on it

rgraber: We could just leave the actual decision about what version to use up to enterprise

rgraber: @robrap who owns the decision about how we do or don't use max constraints?

robrap: My recommendation would be:
1. Make this `3.3.1`.
2. Fix the broken max constraint by simply removing that max constraint.
3. Notify enterprise and let them leave it unconstrained or add in a non-broken max constraint.

rgraber: It sounds like enterprise isn't actually sure if the constraint can safely be removed. @georgebabey

setup.py
timmc-edx: On a different note, I think we'll want to put this into the `load_requirements` function itself to make the semgrep transformation easier. (This isn't performance sensitive, so no real benefit from pre-compilation.) We can also `import re` at the top of the function to make it more self-contained.

setup.py
```diff
@@ -31,25 +33,54 @@ def get_version(*file_paths):
...
-    Load all requirements from the specified requirements files.
+    Load all requirements from the specified requirements files, including any constraints from other files that
+    are pulled in.
     Returns a list of requirement strings.
     """
-    requirements = set()
+    requirements = {}
+    constraint_files = set()
+
+    def add_version_constraint_or_raise(current_line, current_requirements, add_if_not_present):
```
timmc-edx, line 43: I have a minor preference to make this a closure over `requirements` rather than passing it in as an explicit variable, since there's less indirection that way. (If this were lifted out as a separate function so it could be unit tested, I'd probably have a different opinion!)

But I think that's mostly a matter of taste, nothing I'm going to strongly advocate for.

setup.py
timmc-edx: For reference: We have about 130 repos with setup.py files; about 80 have `import os.path` or similar and about 40 have `import re`. So maybe we just need to do some one-time fixup on the import headers? Maybe we can just toss the imports in there and run isort from the PR-generating script to clean it up. :-D

timmc-edx:
Code looks great, just need to find the right repo to test it in. :-)

Note for future work: When we get to the point of mass-updating repos, I think we'll want to have the `load_requirements` docstring point to the template source file (not sure where that will be) so that people will know to update that one instead of the local one in the individual repo.

robrap:
How do we lan on reconciling “mass update” with the fact that we hit an exceptional case on our first repo? What will be the process of detecting and marking exceptions?

timmc-edx:
In a previous comment I suggested that each PR come with a diff of the effective `install_requires` values. If there's a difference, we can have a different PR body than for clean diffs, such as an a checkbox saying "I've confirmed that the differences here are acceptable" and some text about using an appropriate version bump.

PR #284, opened 2022-02-14
feat!: drop support for Django 2.2, 3.0 and 3.1

iamsobanjaved:
**Merge checklist:**

**Post merge:**


.github/workflows/ci.yml
```diff
@@ -49,7 +48,7 @@ jobs:
       run: tox
 
     - name: Run Coverage
-      if: matrix.python-version == '3.8' && matrix.toxenv=='django22'
+      if: matrix.python-version == '3.8' && matrix.toxenv=='django32'
```
aht007: Coverage wouldn't run if we do it this way. It should be either `django32-data` or `django32-reporting`.

.github/workflows/ci.yml
```diff
@@ -15,8 +15,7 @@ jobs:
       matrix:
         os: [ubuntu-20.04]
         python-version: ['3.8']
-        toxenv: [django22-data, django30-data, django31-data, django32-data, django22-reporting,
-                 django30-reporting, django31-reporting, django32-reporting, quality]
+        toxenv: [quality, django32-data, django32-reporting]
```
aht007, line 18: Are we having trouble adding `Django40` testing to this repo?

enterprise_data/__init__.py
```diff
@@ -2,6 +2,6 @@
 Enterprise data api application. This Django app exposes API endpoints used by enterprises.
 """
 
-__version__ = "3.3.0"
+__version__ = "4.0.0"
```
aht007, line 5: It should be a minor version bump since we are just dropping support for older versions.

.github/workflows/ci.yml
iamsobanjaved: Yes, there were failures, added a ticket for that here: https://openedx.atlassian.net/browse/BOM-3292

enterprise_data/__init__.py
iamsobanjaved: Isn't It a breaking change?  we are dropping support for Django 2.2 (LTS version)

requirements/pip.txt
```diff
@@ -4,11 +4,11 @@
...
 #
-wheel==0.37.0
+wheel==0.37.1
     # via -r requirements/pip.in
 
 # The following packages are considered to be unsafe in a requirements file:
 pip==20.0.2
     # via -r requirements/pip.in
-setuptools==58.3.0
+setuptools==60.9.0
```
aht007: We should advertise common constraint file in pip.in here since we have constrained `setuptools` in common constraint.

enterprise_data/__init__.py
aht007: We had a conversation on this within the team. You can see that [here](https://github.com/openedx/edx-ora2/pull/1807#discussion_r804624520). So we are following this approach across the repos

.github/workflows/ci.yml
aht007: We already have a ticket in the backlog. If its about the Middleware issue then you can add that here instead.
https://openedx.atlassian.net/browse/BOM-3282

requirements/base.txt
```diff
@@ -45,23 +38,22 @@ cffi==1.15.0
...
     #   -r requirements/reporting.in
-    #   azure-storage-blob
     #   django-fernet-fields
     #   paramiko
     #   pgpy
     #   pyjwt
     #   pyopenssl
     #   snowflake-connector-python
-django==2.2.24
+django==3.2.12
```
awais786, line 56: Major ver bump

awais786 (approved):

aht007:
@iamsobanjaved Please also update the trove classifiers before merging.

UsamaSadiq (approved):

awais786 (approved):

PR #293, opened 2022-03-15
fix: add rate-limiting retry for enterprise requests

alangsto:
## [MST-1304 ](https://openedx.atlassian.net/browse/MST-1304)

Every time the [enterprise-reporting jenkins job](https://github.com/edx/jenkins-job-dsl-internal/blob/master/jobs/tools-edx-jenkins.edx.org/createEnterpriseDataReportingJobs.groovy) runs, it creates spikes of errors for prod analytics api. This is because the jenkins job makes many calls to an[ endpoint that is rate limited](https://github.com/openedx/edx-enterprise-data/blob/22d05deeac3b25cd73bd659e0a92840ea89c683b/enterprise_data/clients.py#L74), and eventually errors out when that rate limit is reached. If we know that we may get a rate limiting error, we should catch that exception and retry after the appropriate amount of time has passed.

**Merge checklist:**
- [x] `make static` has been run to update webpack bundling if any static content was updated
- [x] Translations updated (see docs/internationalization.rst but also this isn't blocking for merge atm)

**Post merge:**
- [x] Tag pushed and a new [version](https://github.com/edx/edx-enterprise-data/releases) released
- [x] After versioned build finishes in [Travis](https://travis-ci.org/github/edx/edx-enterprise-data), verify version has been pushed to [PyPI](https://pypi.org/project/edx-enterprise-data/)
- [x] PR created in [edx-analytics-data-api](https://github.com/edx/edx-analytics-data-api) to upgrade dependencies (including edx-enterprise-data)

### DIFF
--- a/enterprise_reporting/clients/__init__.py
+++ b/enterprise_reporting/clients/__init__.py
@@ -6,8 +6,10 @@ import os
 from datetime import datetime
 from functools import wraps
 from urllib.parse import parse_qs, urlparse
+import time
 
 from edx_rest_api_client.client import EdxRestApiClient
+from edx_rest_api_client.exceptions import HttpClientError
 
 
 class EdxOAuth2APIClient:
@@ -129,7 +131,16 @@ def traverse_pagination(response, endpoint):
     next_page = response.get('next')
     while next_page:
         querystring = parse_qs(urlparse(next_page).query, True)
-        response = endpoint.get(**querystring)
+        try:
+            response = endpoint.get(**querystring)
+        except HttpClientError as exc:
+            # If we get HTTP429, we should pause for a period of time before re-requesting one more time.
+            # This status code is expected because of rate limiting on the endpoint
+            if exc.response.status_code == 429:
+                time.sleep(5)
+                response = endpoint.get(**querystring)
+            else:
+                raise exc
         results += response.get('results', [])
         next_page = response.get('next')
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,10 @@ Unreleased
 ----------
 
 =========================
+[4.2.1] - 2022-03-15
+---------------------
+  * Added error handling for rate limit exceptions
+
 [4.2.0] - 2022-03-15
 ---------------------
   * Removed currently broken admin url inclusion from enterprise-data.
--- a/enterprise_data/__init__.py
+++ b/enterprise_data/__init__.py
@@ -2,6 +2,6 @@
 Enterprise data api application. This Django app exposes API endpoints used by enterprises.
 """
 
-__version__ = "4.2.0"
+__version__ = "4.2.1"
 
 default_app_config = "enterprise_data.apps.EnterpriseDataAppConfig"  # pylint: disable=invalid-name


alangsto:
One thing that I've been trying to figure out is that there is error handling for function that eventually calls the endpoint where we get the rate limiting error: https://github.com/openedx/edx-enterprise-data/blob/22d05deeac3b25cd73bd659e0a92840ea89c683b/enterprise_reporting/send_enterprise_reports.py#L34-L37 It's not working as expected though.

alangsto:
There's also not an existing test framework for sending reports, so I've been trying to see how to add a test for the error handling that I added.

enterprise_reporting/clients/__init__.py
```diff
@@ -129,7 +131,16 @@ def traverse_pagination(response, endpoint):
...
     while next_page:
         querystring = parse_qs(urlparse(next_page).query, True)
-        response = endpoint.get(**querystring)
+        try:
+            response = endpoint.get(**querystring)
+        except HttpClientError as exc:
+            # If we get HTTP429, we should pause for a minute before re-requesting one more time. This status code
+            # is expected because of rate limiting on the endpoint
+            if exc.response.status_code == 429:
+                time.sleep(65)
```
christopappas: personal preference, but, the duration of the sleep feels like it should be a captured in settings var and then that var be used here

65 seems like a sane enough to me, but curious if there was anything in particular that landed you on that number?

alangsto: The rate-limiting requests per time is defined here:https://github.com/openedx/edx-platform/blob/db32ff2cdf678fa8edd12c9da76a76eef0478614/lms/envs/common.py#L3226. It's set at 800/minute, so I'm being overly cautious with 65 seconds. 

I'll update the PR with your suggestion to use a settings var, thank you!

macdiesel: I'm wondering if 65 is too much? That 800 is per worker node and there are likely other worker nodes out there that have capacity. I have not investigated the error myself but do you see hundreds or thousands of these in new relic? If not it's likely that that other nodes are taking up the capacity for the nodes that have been rate limited. 

Another option here might be to add a timeout but also then rework the alarm to not include these rate limiting events. Do you really care when this happens? We recently started ignoring a few exceptions in our error logging in new relic that did not have meaning to us. Perhaps you may also want to consider ignoring these rate limited alerts so you do not get paged.

I think that 65 is too high of a number. I think this could lead to longer than desired run times and honestly the rate limiting is doing it's job in preventing the server from falling over. How about setting this to 5, or 10?

alangsto: The error definitely doesn't happen hundreds or thousands of times an hour - it looks like it averages ~30 per hour. We don't really care that this happens (especially now that this is due to an enterprise jenkins job as opposed to an error with the insights health check or something like that). 

I still think it would be beneficial to add a timeout here because the rate limit error is now expected, and I'm not sure what the effects of this unhandled exception are downstream for enterprise reporting (although it probably isn't that significant given that it runs every hour). 

I can reduce the time to 5 seconds 👍

macdiesel (approved):

macdiesel:
👍 lets start small and adjust up if necessary. Thank you.

PR #315, opened 2022-06-14
feat: add EnterpriseOffer model and viewset, make upgrade

long74100:
Model matches table show here: https://github.com/edx/prefect-flows/pull/160/files.

**Merge checklist:**

**Post merge:**


enterprise_data/tests/test_utils.py
```diff
@@ -231,6 +231,49 @@ def set_fields_according_to_consent(
...
+    offer_id = factory.lazy_attribute(
+        lambda x: FAKER.random_int(min=1, max=999999)
+    )
+    offer_type = 'Site'
+    status = 'Open'
+    enterprise_customer_uuid = str(FAKER.uuid4())
+    max_discount = factory.lazy_attribute(
+        lambda x: FAKER.pyfloat(right_digits=2, min_value=10000, max_value=100000)
+    )
+    discount_value = factory.lazy_attribute(
```
christopappas: nit: this can be an absolute number, in addition to a percentage. it's probably fine as a test factory, but calling out in case you think it's worth it to add functionality for absolute now 🤷

enterprise_data/models.py
```diff
@@ -197,6 +197,44 @@ def __repr__(self):
...
+    enterprise_customer_uuid = models.UUIDField(db_index=True, null=False)
+    enterprise_name = models.CharField(max_length=255, null=True)
+    total_discount = models.FloatField(null=True)
+    max_discount = models.FloatField(null=True)
+    discount_value = models.FloatField(null=True)
+    discount_type = models.CharField(max_length=255, null=True)
+    sum_amount_learner_paid = models.FloatField(null=True)
+    sum_course_price = models.FloatField(null=True)
+    course_price_covered_by_offer = models.FloatField(null=True)
+    percent_utilized = models.FloatField(null=True)
```
christopappas: took a peek at the warehourse PR, and had a quick Q:

`total_discount ` looked to be `total_discount_ecommerce` in the warehouse PR

also, `percent_utilized` and `course_price_covered_by_offer` don't appear in the warehouse PR (though a thing called `percent_of_offer_spent` exists). should these be properties on this model and not model fields, or is there another field that needs to be exposed via the warehouse PR?

long74100: I'd have to update according to the changes made in the other pr.

enterprise_data/models.py
```diff
@@ -197,6 +197,44 @@ def __repr__(self):
...
+
+    class Meta:
+        app_label = 'enterprise_data'
+        db_table = 'enterprise_offer'
+        verbose_name = _("Enterprise Offer")
+        verbose_name_plural = _("Enterprise Offers")
+
+    offer_id = models.PositiveIntegerField(primary_key=True)
+    offer_type = models.CharField(max_length=255, null=True)
+    status = models.CharField(max_length=255, null=True)
```
adamstankiewicz: What are the statuses an enterprise offer might be? Will be helpful to know for color-coordinating the status badge in the Admin Portal UI.

enterprise_data/models.py
```diff
@@ -197,6 +197,44 @@ def __repr__(self):
...
+    enterprise_customer_uuid = models.UUIDField(db_index=True, null=False)
+    enterprise_name = models.CharField(max_length=255, null=True)
```
adamstankiewicz, line 215: What situations won't we have the `enterprise_name` here (curious about the `null=True`, especially if this field is always included in the `__str__` below)

enterprise_data/models.py
long74100: 'Open' and 'Consumed'

enterprise_data/tests/test_utils.py
long74100: decided to skip Absolute since it's for test data and we probably won't use absolute offers.

enterprise_data/models.py
long74100: I believe every field is nullable by convention since the data is loaded in. It shouldn't be actually be null.

CHANGELOG.rst
```diff
@@ -16,6 +16,10 @@ Unreleased
 
 =========================
 
+[4.2.10] - 2022-06-11
```
iloveagent57: Nit: probably worthy of a 4.3.0

enterprise_data/management/commands/create_enterprise_offer.py
```diff
@@ -0,0 +1,39 @@
+"""
+management command for creating dummy enterprise offers.
+"""
+
+
+import sys
```
iloveagent57, line 6: You could `# pragma: no cover` this whole file to get rid of those codecov checks below.

enterprise_data/migrations/0029_enterpriseoffer.py
```diff
@@ -0,0 +1,39 @@
...
+    dependencies = [
+        ('enterprise_data', '0028_enterpriselearnerenrollment_offer_id'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='EnterpriseOffer',
+            fields=[
+                ('offer_id', models.PositiveIntegerField(primary_key=True, serialize=False)),
+                ('enterprise_customer_uuid', models.CharField(db_index=True, max_length=32)),
```
iloveagent57: I think if you make this a UUIDField you won't have to worry about manually removing those dashes.

iloveagent57 (approved):
Looks great, couple small suggestions.

PR #340, opened 2022-10-12
build: run coverage for all unit test checks

UsamaSadiq:
### Description
- In the PR https://github.com/openedx/edx-enterprise-data/pull/299 the coverage check was refactored to run on only `django-data` tests but we need to cover `django-reporting` tests as well so updating the workflow to run the coverage check for both `django-data` and `django-reportinig`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
       run: tox
 
     - name: Run Coverage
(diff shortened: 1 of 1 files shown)


UsamaSadiq:
@muhammad-ammar @jmbowman arbi-bom doesn't have permissions to review & merge in this repo. Please have a look to verify if this approach would work according to our needs.

muhammad-ammar (approved):
Awesome. Thank you for fixing this coverage issue.

PR #341, opened 2022-10-19
refactor: update cache key and remove session dependency

muhammad-ammar:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise_data/management/commands/tests/test_create_enterprise_learner_enrollment_lpr_v1.py
+++ b/enterprise_data/management/commands/tests/test_create_enterprise_learner_enrollment_lpr_v1.py
@@ -35,12 +35,12 @@ class TestCreateEnterpriseLearnerEnrollmentCommand(TestCase):
             enterprise_customer_uuid=args[0]
         )
         assert enterprise_learner_enrollment.count() == 1
-        assert enterprise_learner_enrollment[0].progress_status is not None
-        assert enterprise_learner_enrollment[0].letter_grade is not None
-        assert enterprise_learner_enrollment[0].enterprise_user_id is not None
-        assert enterprise_learner_enrollment[0].user_username is not None
-        assert enterprise_learner_enrollment[0].user_email is not None
-        assert enterprise_learner_enrollment[0].enterprise_user is not None
+        assert enterprise_learner_enrollment[0].progress_status is None
+        assert enterprise_learner_enrollment[0].letter_grade is None
+        assert enterprise_learner_enrollment[0].enterprise_user_id is None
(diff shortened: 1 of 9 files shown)


enterprise_data/management/commands/tests/test_create_enterprise_learner_enrollment_lpr_v1.py
```diff
@@ -57,13 +57,13 @@ def test_create_enterprise_learner_enrollment_lpr_v1_with_dsc_enabled(self):
...
-        assert enterprise_learner_enrollment[0].user_username is None
-        assert enterprise_learner_enrollment[0].enterprise_user is None
-        assert enterprise_learner_enrollment[0].user_email is None
+        assert enterprise_learner_enrollment[0].letter_grade is not None
+        assert enterprise_learner_enrollment[0].last_activity_date is not None
+        assert enterprise_learner_enrollment[0].progress_status is not None
+        assert enterprise_learner_enrollment[0].enterprise_user_id is not None
+        assert enterprise_learner_enrollment[0].user_username is not None
+        assert enterprise_learner_enrollment[0].enterprise_user is not None
+        assert enterprise_learner_enrollment[0].user_email is not None
```
HammadAhmadWaqas, line 66: Why do we have to revert the checks for this test and the test above?

HammadAhmadWaqas (approved):
Just 1 comment, Otherwise looks good.

enterprise_data/management/commands/tests/test_create_enterprise_learner_enrollment_lpr_v1.py
muhammad-ammar: Due to https://github.com/openedx/edx-enterprise-data/pull/341/files#diff-6cc15faa047fb8c27baf0f209d3a39628c40fd0be410729e877d815494681d30R231

enterprise_data/tests/test_utils.py
```diff
@@ -229,7 +228,7 @@ def set_fields_according_to_consent(
             'user_country_code',
             'user_username',
         ]
-        if create and obj.is_consent_granted:
+        if create and not obj.is_consent_granted:
```
HammadAhmadWaqas, line 231: So there was a bug in the existing logic?

muhammad-ammar: yes.

muhammad-ammar: The logic is to set couple of fields to `None` if `is_consent_granted` is `False`. But the above logic in the factory class is doing the opposite.

PR #370, opened 2023-06-14
Allow querying of offers by either new style UUIDs or old style enterprise ID numbers.

macdiesel:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise_data/api/v1/serializers.py
+++ b/enterprise_data/api/v1/serializers.py
@@ -2,13 +2,9 @@
 Serializers for enterprise api v1.
 """
 
-import logging
-
 from rest_framework import serializers
 
 from enterprise_data.models import EnterpriseLearner, EnterpriseLearnerEnrollment, EnterpriseOffer
-
-LOGGER = logging.getLogger(__name__)
 
 
 class EnterpriseLearnerEnrollmentSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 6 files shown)


enterprise_data/tests/api/v1/test_serializers.py
```diff
@@ -50,54 +50,46 @@ class TestEnterpriseOfferSerializer(APITransactionTestCase):
     """
 
     @ddt.data(
-        ('787f0fa7-f89a-4267-9e2b-fcd299e83918', '787f0fa7-f89a-4267-9e2b-fcd299e83918'),
+        ('787f0fa7-f89a-4267-9e2b-fcd299e83918', '787f0fa7f89a42679e2bfcd299e83918'),
+        ('56a779a5-21ec-4711-830f-6654d3faaa5f', '56a779a521ec4711830f6654d3faaa5f'),
+    )
+    @ddt.unpack
+    def test_deserialize_offer_id_valid(self, offer_id_request, offer_id_db):
```
pwnage101, line 58: This updated test no longer includes success cases for integer offer_ids in the request.  We should probably add the following:
* 123456789
* 1

pwnage101 (approved):
otherwise, LGTM!

enterprise_data/tests/api/v1/test_serializers.py
```diff
@@ -52,6 +52,7 @@ class TestEnterpriseOfferSerializer(APITransactionTestCase):
     @ddt.data(
         ('787f0fa7-f89a-4267-9e2b-fcd299e83918', '787f0fa7f89a42679e2bfcd299e83918'),
         ('56a779a5-21ec-4711-830f-6654d3faaa5f', '56a779a521ec4711830f6654d3faaa5f'),
+        ('12345', 12345),
```
pwnage101, line 55: the right hand side should be a string, since this is a CharField, right?

macdiesel: Serializer disagreed. Ultimately I don't think it'll make a difference but something we can consider cleaning up.

enterprise_data/api/v1/serializers.py
```diff
@@ -63,35 +59,38 @@ class Meta:
...
+                ret['offer_id'] = offer_id
+                return ret
+
+            else:
+                raise serializers.ValidationError("requested offer_id neither a valid integer nor UUID.")
+
+        if len(ret['offer_id']) < 10:  # All ecommerce offer_ids are at < 1 million.
+            try:
+                ret['offer_id'] = int(ret['offer_id'])
+                return ret
```
pwnage101, line 89: this is causing an attempt to store an integer into the database.  It'll work as intended, but cleaner to just use int() as a test:
```suggestion
                int(ret['offer_id'])
                return ret
```

PR #397, opened 2023-09-07
Revert "Revert "feat: ENT-7519 Added EnterpriseSubsidyBudget model""

muhammad-ammar:
Reverts openedx/edx-enterprise-data#395


enterprise_data/models.py
```diff
@@ -323,6 +324,57 @@ def __repr__(self):
...
+
+    objects = EnterpriseReportingModelManager()
+
+    class Meta:
+        app_label = 'enterprise_data'
+        db_table = 'ent_subsidy_access_policy_aggregates'
+        verbose_name = _("Enterprise Subsidy Budget")
+        verbose_name_plural = _("Enterprise Subsidy Budgets")
+
+    id = models.CharField(db_index=True, primary_key=True, max_length=32)
```
irfanuddinahmad, line 341: Can this hash be more than 32?

irfanuddinahmad (approved):

enterprise_data/models.py
muhammad-ammar: In our warehouse, we use md5 hash and according to [docs](https://docs.snowflake.com/en/sql-reference/functions/md5#returns), it will be a 32 character string

PR #428, opened 2024-02-07
feat: stream csv downloads

muhammad-ammar:
**JIRA:** https://2u-internal.atlassian.net/browse/ENT-8301

**Dependency:** https://github.com/openedx/frontend-app-admin-portal/pull/1167

**Merge checklist:**

**Post merge:**


enterprise_data/api/v1/views.py
```diff
@@ -140,6 +143,31 @@ def get_queryset(self):
...
+        """
+        queryset = self.filter_queryset(self.get_queryset())
+
+        page = self.paginate_queryset(queryset)
+        if page is not None:
+            serializer = self.get_serializer(page, many=True)
+            return self.get_paginated_response(serializer.data)
+
+        def data_gen(queryset):
+            paginator = Paginator(queryset, per_page=10000)
```
iloveagent57: small suggestion: make the `10000` value come from settings, so we can tune it up or down without deployment.

enterprise_data/api/v1/views.py
```diff
@@ -140,6 +143,31 @@ def get_queryset(self):
...
+    def list(self, request, *args, **kwargs):
+        """
+        Override the list method to handle streaming CSV download.
+        """
+        queryset = self.filter_queryset(self.get_queryset())
+
+        page = self.paginate_queryset(queryset)
+        if page is not None:
+            serializer = self.get_serializer(page, many=True)
+            return self.get_paginated_response(serializer.data)
```
iloveagent57: Do we want this here? Should the block of code that conditionally renders the CSV come first?

enterprise_data/api/v1/views.py
```diff
@@ -140,6 +143,31 @@ def get_queryset(self):
             TieredCache.set_all_tiers(cache_key, enrollments, DEFAULT_LEARNER_CACHE_TIMEOUT)
             return enrollments
 
+    def list(self, request, *args, **kwargs):
+        """
+        Override the list method to handle streaming CSV download.
+        """
```
iloveagent57, line 141: One idea for rollout: introduce a feature flag, where if the flag is off, this method can probably just return `super().list(...)`. And if it's on, it can do all the new stuff you've introduced.

enterprise_data/renderers.py
```diff
@@ -0,0 +1,20 @@
+from rest_framework_csv.renderers import CSVStreamingRenderer
+
+
+class EnrollmentsCSVRenderer(CSVStreamingRenderer):
+    header = [
```
iloveagent57, line 17: You could probably do something like
```python
header = [field.name for field in EnterpriseLearnerEnrollment._meta.get_fields()]
```

enterprise_data/api/v1/views.py
iloveagent57: Also, here's an example from SO that might give you an idea for how to structure this code a little differently: https://stackoverflow.com/a/65564367 It might help you simplify this a little bit.

enterprise_data/api/v1/views.py
```diff
@@ -140,6 +143,31 @@ def get_queryset(self):
...
+            return self.get_paginated_response(serializer.data)
+
+        def data_gen(queryset):
+            paginator = Paginator(queryset, per_page=10000)
+            for page_number in paginator.page_range:
+                enrollments = paginator.page(page_number)
+                serializer = self.get_serializer(enrollments, many=True)
+                yield serializer.data
+
+        if self.request.query_params.get('data') == 'csv':
```
iloveagent57: Why is this change necessary? Does requesting `...enrollments.csv` no longer work?

iloveagent57: You might be able to do something like this if you want to switch behavior based on whether `.csv` is being requested: https://www.django-rest-framework.org/api-guide/renderers/#varying-behavior-by-media-type

muhammad-ammar: > Why is this change necessary? Does requesting `...enrollments.csv` no longer work?

Unfortunately, yes. I really want to avoid this. I will give it another look.

muhammad-ammar:
@iloveagent57 Bundle of the thanks for the feedback. This work is in progress. I wanted to get your thoughts on the overall approach. I will address all the feedback. 

One question, Things are clear regarding streaming csv. How about chunking the DB reads?

enterprise_data/api/v1/views.py
muhammad-ammar: Requests like `enrollments.csv` will now work

enterprise_data/api/v1/views.py
muhammad-ammar: What do you say if enable/disable old/new functionality based on a query param passed from admin-portal?

iloveagent57: Yeah, sure, that's a good idea too.

enterprise_data/api/v1/views.py
```diff
@@ -140,6 +146,30 @@ def get_queryset(self):
...
+
+    def _stream_serialized_data(self):
+        """
+        Stream the serialized data.
+        """
+        queryset = self.filter_queryset(self.get_queryset())
+        serializer = self.get_serializer_class()
+        paginator = Paginator(queryset, per_page=settings.ENROLLMENTS_PAGE_SIZE)
+        for page_number in paginator.page_range:
+            yield from serializer(paginator.page(page_number).object_list, many=True).data
```
iloveagent57: Awesome, nice change to clean this up.

iloveagent57 (approved):
The code looks good!  Have you been able to test this out on a fake/large data set?

muhammad-ammar:
> The code looks good! Have you been able to test this out on a fake/large data set?

Yes, I am using the dataset for https://portal.edx.org/national-university-singapore/admin/learners for local testing. 

Results with reading chunks of 10000 records from DB

With streaming 
* CSV Size: 76.8 MB
* CSV Rows: 127500
* Average: 33 seconds

Without streaming
* CSV Size: 76.8 MB
* CSV Rows: 127500
* Average:  28 seconds

PR #480, opened 2024-08-16
refactor: Add logs and time measurements for different code blocks

muhammad-ammar:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise_data/api/v1/views/analytics_enrollments.py
+++ b/enterprise_data/api/v1/views/analytics_enrollments.py
@@ -1,5 +1,6 @@
 """Advance Analytics for Enrollments"""
 from datetime import datetime
+from logging import getLogger
 
 from edx_rbac.decorators import permission_required
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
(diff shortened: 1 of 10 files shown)


enterprise_data/api/v1/views/analytics_enrollments.py
```diff
@@ -112,49 +129,55 @@ def get(self, request, enterprise_uuid):  # lint-amnesty, pylint: disable=incons
...
-                "top_subjects_by_enrollments": self.construct_top_subjects_by_enrollments(
-                    enrollments_df.copy(),
-                    start_date,
-                    end_date,
-                ),
-            }
+            with timer('construct_enrollment_all_stats'):
+                data = {
+                    "enrollments_over_time": self.construct_enrollments_over_time(
+                        enrollments_df.copy(),
```
saleem-latif, line 138: I think we should remove calls to `copy` here.

jajjibhai008 (approved):
LGTM 👍

enterprise_data/utils.py
```diff
@@ -68,6 +69,21 @@ def wrapper(*args, **kwargs):
...
+    """
+    Context manager to measure the time taken by a block of code.
+
+    Arguments:
+        prefix (str): The prefix to print in the log.
+    """
+    start = time.time()
+    yield
+    end = time.time()
+    difference = end - start
```
saleem-latif, line 83: nit: We can remove `end` and directly call `time.time()  - start`

saleem-latif (approved):

enterprise_data/api/v1/views/analytics_enrollments.py
muhammad-ammar: copy is needed here because every function mutates the original data frame.

PR #500, opened 2024-09-20
feat: Transformed pandas processing into SQL for leaderboard

irfanuddinahmad:
**Merge checklist:**

**Post merge:**


enterprise_data/admin_analytics/database/queries/fact_engagement_admin_dash.py
```diff
@@ -19,3 +19,43 @@ def get_learning_hours_and_daily_sessions_query():
...
+            SELECT
+                Engagement.email,
+                learning_time_hours,
+                sessions,
+                average_session_length,
+                course_completions
+            FROM Engagement
+            LEFT JOIN Completions
+            ON Engagement.email = Completions.email
+            ORDER BY learning_time_hours desc, sessions desc, course_completions desc
```
muhammad-ammar: Facing the below error locally on read replica when I execute the query in mysql shell.

```
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '== 0 THEN 0.0
            ELSE ROUND(learning_time_hours / sessions::float, 1)
```

enterprise_data/api/v1/views/analytics_leaderboard.py
```diff
@@ -36,61 +30,17 @@ class AdvanceAnalyticsLeaderboardView(APIView):
...
-
-        leaderboard_df = engage.merge(completions, on="email", how="left")
-        leaderboard_df = leaderboard_df.sort_values(
-            by=["learning_time_hours", "daily_sessions", "course_completions"],
-            ascending=[False, False, False],
-        )
-
+        response_type = request.query_params.get('response_type', ResponseType.JSON.value)
+        leaderboard = FactEngagementAdminDashTable().get_leaderboard(enterprise_uuid, start_date, end_date)
+        leaderboard_df = pd.DataFrame(leaderboard)
```
muhammad-ammar: Why are we again converting the results to dataframe?

muhammad-ammar: I believe we can convert the below logic into python without using pandas dataframe. We have planned to remove the pandas once we convert all the code to mysql queries.

```
# move the aggregated row with email 'null' to the end of the table
idx = leaderboard_df.index[leaderboard_df['email'] == 'null']
leaderboard_df.loc[idx, 'email'] = 'learners who have not shared consent'
leaderboard_df = pd.concat([leaderboard_df.drop(idx), leaderboard_df.loc[idx]])
```

irfanuddinahmad: pandas is python also!

muhammad-ammar: >  We can convert the below logic into python **without using pandas dataframe**.

irfanuddinahmad: I think pandas might be more efficient in this processing ...

muhammad-ammar (approved):

PR #521, opened 2024-10-14
ENT-9622: Added caching for API endpoints related to advanced analytics.

saleem-latif:
__Jira Ticket:__ [ENT-9622](https://2u-internal.atlassian.net/browse/ENT-9622)

__Description:__
This PR adds caching for the REST API endpoints related to advanced analytics. Currently, the timeout is set to 1 hour but we can adjust it after observing the current implementation on production.

**Merge checklist:**

**Post merge:**


enterprise_data/cache/decorators.py
```diff
@@ -0,0 +1,42 @@
...
+    def inner_decorator(func):
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            """
+            Wrapper function to cache the result of the function.
+            """
+            cache_key = cache.get_key(func.__name__, *args, **kwargs)
+            cached_response = cache.get(cache_key)
+            if cached_response.is_found:
+                LOGGER.info("[ANALYTICS]: Cache hit for key: (%s)", (func.__name__, args, kwargs))
```
muhammad-ammar, line 34: Don't we need to have a `%s` format specifiers for each argument like below?

```python
LOGGER.info("[ANALYTICS]: Cache hit for key: (%s, %s, %s)", func.__name__, args, kwargs)
```

enterprise_data/cache/decorators.py
```diff
@@ -0,0 +1,42 @@
...
+                return cached_response.value
+
+            LOGGER.info("[ANALYTICS]: Cache miss for key: (%s)", (func.__name__, args, kwargs))
```
muhammad-ammar, line 37: Same as above

enterprise_data/cache/decorators.py
```diff
@@ -0,0 +1,42 @@
...
+    Arguments:
+        timeout (int): Cache timeout in seconds.
+
+    Returns:
+        (function): Decorator function.
+    """
+
+    def inner_decorator(func):
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
```
muhammad-ammar, line 27: Currently this decorator is written to work only for class methods, can we modify it to work with class and non-class methods?

muhammad-ammar (approved):
LGTM

enterprise_data/cache/decorators.py
saleem-latif: Nope, I am passing in a tuple of all values so a single %s is the right approach.

enterprise_data/cache/decorators.py
saleem-latif: Since, our need is to use it just for class or instance methods. I do not think adding unnecessary complexity would be a good idea.

saleem-latif: I have already mentioned it in the doc string that this decorator would work only for class/instance methods

PR #571, opened 2025-03-05
feat: support flex groups in csv report on LPR

jajjibhai008:
**Description**

JIRA -> [ENT-10025](https://2u-internal.atlassian.net/browse/ENT-10025)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise_data/api/v1/serializers.py
+++ b/enterprise_data/api/v1/serializers.py
@@ -6,6 +6,7 @@ from uuid import UUID
 from rest_framework import serializers
 
 from enterprise_data.admin_analytics.constants import ResponseType
+from enterprise_data.cache.decorators import cache_it
 from enterprise_data.models import (
     EnterpriseAdminLearnerProgress,
     EnterpriseAdminSummarizeInsights,
@@ -26,6 +27,8 @@ class EnterpriseLearnerEnrollmentSerializer(serializers.ModelSerializer):
     course_api_url = serializers.SerializerMethodField()
     enterprise_user_id = serializers.SerializerMethodField()
     total_learning_time_hours = serializers.SerializerMethodField()
+    enterprise_flex_group_name = serializers.SerializerMethodField()
+    enterprise_flex_group_uuid = serializers.SerializerMethodField()
 
     class Meta:
         model = EnterpriseLearnerEnrollment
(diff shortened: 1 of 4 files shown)


enterprise_data/api/v1/serializers.py
```diff
@@ -62,6 +66,52 @@ def get_total_learning_time_hours(self, obj):
         """Returns the learners total learning time in hours"""
         return round((obj.total_learning_time_seconds or 0.0)/3600.0, 2)
 
+    @cache_it()
+    def _get_flex_groups(self, obj):
+        """
+        Returns list of tuples containing group (name, uuid) pairs for the learner.
+        This is cached to prevent duplicate database queries.
+        """
+        print('ayaaaaaa')
```
omar-sarfraz: Looks like a log line was pushed mistakenly. Please remove it before merging 👍

enterprise_data/api/v1/serializers.py
```diff
@@ -62,6 +66,52 @@ def get_total_learning_time_hours(self, obj):
...
+            .values_list("enterprise_group_name", "enterprise_group_uuid")
+            .distinct()
+        )
+
+    def get_enterprise_flex_group_name(self, obj):
+        """Returns a comma-separated list of enterprise group names that the learner is associated with"""
+        groups = self._get_flex_groups(obj)
+
+        if not groups:
+            return obj.enterprise_group_name
```
omar-sarfraz, line 99: [Curious] We have removed enterprise_group_name from serializer fields. So, this probably won't return anything. Can we just return an empty string if no groups are found?

enterprise_data/api/v1/serializers.py
```diff
@@ -62,6 +66,52 @@ def get_total_learning_time_hours(self, obj):
...
+
+        # Return comma-separated list of group names (first element of each tuple)
+        return ', '.join(group[0] for group in groups)
+
+    def get_enterprise_flex_group_uuid(self, obj):
+        """Returns a comma-separated list of enterprise group UUIDs that the learner is associated with"""
+        groups = self._get_flex_groups(obj)
+
+        if not groups:
+            return obj.enterprise_group_uuid
```
omar-sarfraz, line 109: Same here

omar-sarfraz (changes requested):

enterprise_data/api/v1/serializers.py
jajjibhai008: This is so because enrollments data also contain the flex group info in which that specific learner is attached, so if no group returns from the membership table, then we will use this single group that is coming in the enrollment record

enterprise_data/api/v1/serializers.py
jajjibhai008: answers above.

omar-sarfraz (approved):
LGTM

enterprise_data/api/v1/serializers.py
```diff
@@ -62,6 +66,51 @@ def get_total_learning_time_hours(self, obj):
...
+    @cache_it()
+    def _get_flex_groups(self, obj):
+        """
+        Returns list of tuples containing group (name, uuid) pairs for the learner.
+        This is cached to prevent duplicate database queries.
+        """
+        enterprise_user_id = obj.enterprise_user_id
+
+        if not enterprise_user_id:
+            return []
```
muhammad-ammar, line 78: For conditions like this, you can use [walrus operator](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions) `:=`. For example

```python
if not (enterprise_user_id := obj.enterprise_user_id):
    return []
```

muhammad-ammar (approved):

PR #581, opened 2025-04-07
feat: Added the ability to filter enrollments by group_uuid in the en…

jajjibhai008:
…terprise completions API

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise_data/admin_analytics/database/filters/fact_completion_admin_dash.py
+++ b/enterprise_data/admin_analytics/database/filters/fact_completion_admin_dash.py
@@ -1,0 +1,36 @@
+"""
+Query filters for enrollments table.
+"""
+from enterprise_data.admin_analytics.database.filters.base import BaseFilter
+from enterprise_data.admin_analytics.database.query_filters import BetweenQueryFilter, EqualQueryFilter
+
+
+class FactCompletionAdminDashFilters(BaseFilter):
+    """
+    Query filters for completions data in enrollments table.
+    """
+    @staticmethod
+    def passed_date_range_filter(
+            start_date_params_key: str, end_date_params_key: str
+    ) -> BetweenQueryFilter:
+        """
+        Filter by passed date to be in the given range.
+
+        Arguments:
+            start_date_params_key (str): The start date key against which value will be passed in the query.
+            end_date_params_key (str): The end date key against which value will be passed in the query.
+        """
+        return BetweenQueryFilter(
+            column='passed_date',
+            range_placeholders=(start_date_params_key, end_date_params_key),
+        )
+
+    @staticmethod
+    def has_passed_filter() -> EqualQueryFilter:
+        """
+        Filter by has passed with fixed value 1.
+        """
+        return EqualQueryFilter(
+            column='has_passed',
+            value=1
+        )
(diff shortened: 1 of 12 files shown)


enterprise_data/admin_analytics/database/tables/fact_enrollment_admin_dash.py
```diff
@@ -92,6 +93,43 @@ def __get_common_query_filters(
 
         return query_filters, params
 
+    def __get_common_query_filters_for_completion(
+            self, enterprise_customer_uuid: UUID,
+            group_uuid: Optional[UUID],
+            start_date: date,
+            end_date: date,
+    ) -> (QueryFilters, dict):
```
muhammad-ammar: @jajjibhai008 @saleem-latif 

I think the expression (QueryFilters, dict) is interpreted as a regular tuple, not a type. We need to use Tuple from the typing module like below

```python
from typing import Tuple, Optional
from uuid import UUID
from datetime import date

def __get_common_query_filters_for_completion(
    self, 
    enterprise_customer_uuid: UUID,
    group_uuid: Optional[UUID],
    start_date: date,
    end_date: date,
) -> Tuple[QueryFilters, dict]:

```

enterprise_data/admin_analytics/database/filters/fact_completion_admin_dash.py
```diff
@@ -0,0 +1,36 @@
+"""
+Query filters for enrollments table.
+"""
+from enterprise_data.admin_analytics.database.filters.base import BaseFilter
+from enterprise_data.admin_analytics.database.query_filters import BetweenQueryFilter, EqualQueryFilter
+
+
+class FactCompletionAdminDashFilters(BaseFilter):
+    """
+    Query filters for completion table.
```
saleem-latif: I do not think we have a completion table. This could be something like

```
Query filters for completions data in enrollments table.
```

enterprise_data/admin_analytics/database/filters/fact_completion_admin_dash.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    @staticmethod
+    def enterprise_completion_date_range_filter(
```
saleem-latif: A better name would be `passed_date_range_filter`

enterprise_data/admin_analytics/database/tables/fact_enrollment_admin_dash.py
```diff
@@ -232,27 +272,31 @@ def get_enrollment_and_course_count(self, enterprise_customer_uuid: UUID, start_
...
-                'start_date': start_date,
-                'end_date': end_date,
-            }
+            query=self.queries.get_completion_count_query(query_filters),
+            params=query_filter_params,
+            as_dict=False,
         )
-        if not results:
+        # Handle empty results defensively
+        if not results or not results[0] or len(results[0]) == 0:
```
saleem-latif, line 299: Nice 👍🏼

enterprise_data/api/v1/views/enterprise_admin.py
```diff
@@ -113,7 +113,7 @@ def get(self, request, enterprise_id):
             enterprise_id, start_date, end_date,
         )
         completions = FactEnrollmentAdminDashTable().get_completion_count(
-            enterprise_id, start_date, end_date,
+            enterprise_id, None, start_date, end_date,
```
saleem-latif: You need to extract `group_uuid` from `request.GET` and pass it here, similar to what is being done at https://github.com/openedx/edx-enterprise-data/blob/37841b810aab8473e33f9536a959c5c47b41d403/enterprise_data/api/v1/views/analytics_enrollments.py#L55

saleem-latif (changes requested):

saleem-latif (approved):

PR #643, opened 2025-11-18
feat: filter out unenrollment and update filter order in Learner Prog…

sjasti-sonata-svg:
…ress Report

**This PR introduces a new unenrollment filter to the Learner Progress Report (LPR) and updates the ordering of existing filters.

Added Unenrollment option(Enrolled and UnEnrolled) to LPR filters.

### DIFF
--- a/enterprise_data/api/v1/views/enterprise_learner.py
+++ b/enterprise_data/api/v1/views/enterprise_learner.py
@@ -114,6 +114,7 @@ class EnterpriseLearnerEnrollmentViewSet(EnterpriseViewSetMixin, viewsets.ReadOn
         for page_number in paginator.page_range:
             yield from serializer(paginator.page(page_number).object_list, many=True).data
 
+    # pylint: disable=too-many-statements
     def apply_filters(self, queryset):
         """
         Filters enrollments based on query params.
@@ -180,6 +181,10 @@ class EnterpriseLearnerEnrollmentViewSet(EnterpriseViewSetMixin, viewsets.ReadOn
         group_uuid = query_filters.get('group_uuid')
         if group_uuid:
             queryset = self.filter_by_group_uuid(queryset, group_uuid)
+
+        search_enrollment = query_filters.get('search_enrollment')
+        if search_enrollment in ("enrolled", "unenrolled"):
+            queryset = self.filter_search_enrollment(queryset, search_enrollment)
 
         return queryset
 
@@ -302,6 +307,27 @@ class EnterpriseLearnerEnrollmentViewSet(EnterpriseViewSetMixin, viewsets.ReadOn
         """
         created_max = queryset.aggregate(Max('created'))
         return created_max['created__max']
+
+    def filter_search_enrollment(self, queryset, status):
+        """
+        Filter enrollments based on enrollment `status`.
+
+        Args:
+            status (str): Enrollment status to filter by. Can be one of:
+                'enrolled'   : unenrollment_date is NULL (currently enrolled)
+                'unenrolled' : unenrollment_date is NOT NULL (no longer enrolled)
+
+        Returns:
+            QuerySet: Filtered queryset of enrollments.
+        """
+
+        if status == "enrolled":
+            return queryset.filter(unenrollment_date__isnull=True)
+
+        if status == "unenrolled":
+            return queryset.filter(unenrollment_date__isnull=False)
+
+        return queryset
 
     @action(detail=False)
     def overview(self, request, **kwargs):
--- a/enterprise_data/tests/api/v1/test_views.py
+++ b/enterprise_data/tests/api/v1/test_views.py
@@ -13,6 +13,8 @@ from rest_framework import status
 from rest_framework.reverse import reverse
 from rest_framework.test import APITransactionTestCase
 
+from django.utils import timezone
+
 from enterprise_data.api.v1.serializers import EnterpriseOfferSerializer
 from enterprise_data.models import EnterpriseLearnerEnrollment, EnterpriseOffer
 from enterprise_data.tests.factories import (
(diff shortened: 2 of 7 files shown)


enterprise_data/api/v1/views/enterprise_learner.py
```diff
@@ -303,6 +308,17 @@ def get_max_created_date(self, queryset):
...
+        now=timezone.localdate()
+
+        if search_enrollment == 'enrolled':
+            return queryset.filter(
+                Q(unenrollment_date__isnull=True) | Q(unenrollment_date__gt=now)
+            )
+        elif search_enrollment == 'unenrolled':
+            return queryset.filter(unenrollment_date__lte=now)
+        return queryset
+
```
muhammad-ammar: As per my understanding, `unenrollment_date` will never be in the future. So we can simplify the code as shown below. I believe this will also simplify the unit tests.

```python
def filter_search_enrollment(self, queryset, status):
    """
    Filter enrollments based on enrollment `status`.

    Args:
        status (str): Enrollment status to filter by. Can be one of:
            'enrolled'   : unenrollment_date is NULL (currently enrolled)
            'unenrolled' : unenrollment_date is NOT NULL (no longer enrolled)

   Returns:
       QuerySet: Filtered queryset of enrollments.
    """

    if status == "enrolled":
        return queryset.filter(unenrollment_date__isnull=True)

    if status == "unenrolled":
        return queryset.filter(unenrollment_date__isnull=False)

    return queryset

```

enterprise_data/tests/api/v1/test_views.py
```diff
@@ -369,3 +370,150 @@ def test_retrieve_enterprise_admin_insights_no_access(self):
...
+            unenrollment_date=timezone.now(),
+        )
+
+        response = self.client.get(
+            self._get_url(),
+            data={"search_enrollment": "not-valid"},
+        )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json()["count"], 2)
```
muhammad-ammar: Please add a new line at the end of file.

enterprise_data/tests/api/v1/test_views.py
```diff
@@ -369,3 +370,150 @@ def test_retrieve_enterprise_admin_insights_no_access(self):
...
+        self.addCleanup(mocked_get_enterprise_customer.stop)
+
+        self.enterprise_id = 'fd0d9cd4-bc35-45e8-ba35-e73be3fc5a07'
+        self.set_jwt_cookie()
+
+    def tearDown(self):
+        super().tearDown()
+        EnterpriseLearnerEnrollment.objects.all().delete()
+
+    def _get_url(self):
```
muhammad-ammar: I think instead of creating this method, we can declare a variable in setUp as shown below
```python
self.url = reverse(
    'v1:enterprise-learner-enrollment-list',
    kwargs={'enterprise_id': self.enterprise_id}
)
```

enterprise_data/tests/api/v1/test_views.py
```diff
@@ -369,3 +370,150 @@ def test_retrieve_enterprise_admin_insights_no_access(self):
...
+        return reverse(
+            'v1:enterprise-learner-enrollment-list',
+            kwargs={'enterprise_id': self.enterprise_id}
+        )
+
+    # -----------------------------
+    # 1. Test for enrolled learners
+    # -----------------------------
```
muhammad-ammar: The standard practice is to add docstrings in test functions. Please add docstrings in all tests and remove these comments.

For example,
```
def test_filter_enrolled(self):
"""
Test that the enrollments API returns the correct data when filtered by the 'enrolled' status.
"""
```

enterprise_data/tests/api/v1/test_views.py
```diff
@@ -369,3 +370,150 @@ def test_retrieve_enterprise_admin_insights_no_access(self):
...
+            enterprise_customer_uuid=self.enterprise_id,
+            unenrollment_date=None,
+        )
+
+        # Create unenrolled learner (unenrollment_date != NULL)
+        EnterpriseLearnerEnrollmentFactory(
+            enterprise_user=enterprise_learner,
+            enterprise_customer_uuid=self.enterprise_id,
+            unenrollment_date=timezone.now(),
+        )
```
muhammad-ammar: In all tests, we are creating data inside the tests, it would be great if we can move all the test data creation inside the `setUp` method.

enterprise_data/tests/test_utils.py
```diff
@@ -178,7 +183,7 @@ class Meta:
     )
     letter_grade = factory.lazy_attribute(lambda x: ' '.join(FAKER.words(nb=2)).title())
     progress_status = factory.lazy_attribute(lambda x: ' '.join(FAKER.words(nb=2)).title())
-    enterprise_user_id = factory.Sequence(lambda n: n)
+    #enterprise_user_id = factory.Sequence(lambda n: n)
```
muhammad-ammar: Looks like this was left by mistake?

enterprise_data/tests/test_utils.py
```diff
@@ -154,6 +154,10 @@ class EnterpriseLearnerEnrollmentFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = EnterpriseLearnerEnrollment
 
+    enterprise_user = factory.SubFactory(
+        EnterpriseLearnerFactory,
+        enterprise_user_id=factory.Sequence(lambda n: n+1) 
+    )
```
muhammad-ammar: Could you please explain why we need to make changes in this file?

muhammad-ammar (commented):
Good Work. One last thing. For every PR, you also need make changes in [CHANGELOG.rst](https://github.com/openedx/edx-enterprise-data/blob/master/CHANGELOG.rst) and https://github.com/openedx/edx-enterprise-data/blob/master/enterprise_data/__init__.py

Please see https://github.com/openedx/edx-enterprise-data/pull/641 to understand what changes are needed in the files mentioned above.

enterprise_data/tests/test_utils.py
sjasti-sonata-svg: The previous one was giving an error. FOREIGN KEY constraint failed when I was running Unit test cases.

muhammad-ammar (approved):

PR #670, opened 2026-05-15
feat: Add lowest passing grade to LPR and also add caching to course_progress

rgopalrao-sonata-png:
# [ENT-11788] Add Course Passing Grade Enrichment & Course Progress Caching to LPR

## Summary

This PR implements the backend enrichment layer in `edx-enterprise-data` to support a new **Course Passing Grade** column in the Learner Progress Report (LPR) in the Admin Portal. It also introduces a **per-pair caching strategy** for Course Progress data to eliminate redundant Snowflake queries.

> **Note:** The UI column placement and CSV download integration will be addressed in [ENT-11722](https://2u-internal.atlassian.net/browse/ENT-11722).

---

## Problem

Admins reviewing the Learner Progress Report currently lack visibility into the **required passing grade** for each course. Without this information, it is difficult to assess whether a learner's current progress is on track to meet the course's standard. This creates friction when Admins try to make informed decisions about learner outreach.

Additionally, Course Progress data was being fetched from Snowflake on every API request, causing unnecessary repeated queries for data that changes infrequently (roughly once per day).

---

### 1. Course Passing Grade — New Snowflake Source (`SnowflakeCoursePassingGradeSource`)

- Fetches `LOWEST_PASSING_GRADE` from Snowflake's LMS course-overviews table:
  `PROD.LMS.COURSE_OVERVIEWS_COURSEOVERVIEW`
- Batched query per page of enrollments to minimise Snowflake round-trips
- Per-courserun-key caching with:
  - **Positive cache TTL**: `86400` seconds (24 hours) — configurable via `LPR_COURSE_PASSING_GRADE_CACHE_TIMEOUT`
  - **Negative cache TTL**: `3600` seconds (1 hour) — configurable via `LPR_COURSE_PASSING_GRADE_NEGATIVE_CACHE_TIMEOUT`
- Graceful degradation: if Snowflake is unavailable the field is omitted without breaking the LPR response

### 2. Course Progress Caching — Per-Pair Cache (`SnowflakeCourseProgressSource`)

- Cache is keyed on `(user_email, courserun_key)` — only the exact pairs requested by the current page are looked up or stored.
- Snowflake is queried only for cache-missing pairs using a parameterised `(USER_EMAIL, COURSERUN_KEY) IN (…)` filter together with the enterprise UUID — **never the full enterprise table**.
- Cache TTL: `300` seconds (5 minutes) — configurable via `LPR_COURSE_PROGRESS_CACHE_TIMEOUT`
- Caching follows the project-standard `TieredCache` wrapper (`enterprise_data/cache/__init__.py`).

### 3. Snowflake Connection Factory

- Optional `LPR_SNOWFLAKE_WAREHOUSE` and `LPR_SNOWFLAKE_ROLE` settings are forwarded to the Snowflake connector when present.
- Shared `_get_snowflake_connection()` module-level factory removes duplication between the two source classes.

---

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | A new `course_passing_grade` field is exposed in the LPR API response | ✅ This PR |
| 2 | The column is positioned to the left of "Course Progress" in Admin Portal | 🔜 ENT-11722 |
| 3 | The column displays the required passing grade sourced from Snowflake course overviews | ✅ This PR |
| 4 | The field is included in the LPR CSV download | 🔜 ENT-11722 |
| 5 | Course Progress data is cached per `(user_email, courserun_key)` pair to reduce Snowflake query frequency | ✅ This PR |

---

### Required Settings

| Setting | Description |
|---------|-------------|
| `SNOWFLAKE_SERVICE_USER` | Snowflake service account username |
| `SNOWFLAKE_SERVICE_USER_PASSWORD` | Snowflake service account password |

### Optional Settings (with defaults)

| Setting | Default | Description |
|---------|---------|-------------|
| `SNOWFLAKE_ACCOUNT` | `'edx.us-east-1'` | Snowflake account identifier |
| `LPR_SNOWFLAKE_WAREHOUSE` | `None` | Snowflake virtual warehouse (omitted if unset) |
| `LPR_SNOWFLAKE_ROLE` | `None` | Snowflake role (omitted if unset) |
| `LPR_SNOWFLAKE_DATABASE` | `'PROD'` | Snowflake database for LPR internal table |
| `LPR_SNOWFLAKE_SCHEMA` | `'ENTERPRISE'` | Snowflake schema for LPR internal table |
| `LPR_SNOWFLAKE_INTERNAL_TABLE` | `'LEARNER_PROGRESS_REPORT_INTERNAL'` | Internal LPR table name |
| `LPR_COURSE_PROGRESS_CACHE_TIMEOUT` | `300` | Cache TTL for course progress (seconds) |
| `LPR_SNOWFLAKE_LMS_DATABASE` | `'PROD'` | Snowflake database for course overviews |
| `LPR_SNOWFLAKE_LMS_SCHEMA` | `'LMS'` | Snowflake schema for course overviews |
| `LPR_SNOWFLAKE_COURSE_OVERVIEWS_TABLE` | `'COURSE_OVERVIEWS_COURSEOVERVIEW'` | Course overviews table name |
| `LPR_COURSE_PASSING_GRADE_CACHE_TIMEOUT` | `86400` | Positive cache TTL for passing grades (seconds) |
| `LPR_COURSE_PASSING_GRADE_NEGATIVE_CACHE_TIMEOUT` | `3600` | Negative cache TTL for missing passing grades (seconds) |

---

## Files Changed

| File | Change |
|------|--------|
| `enterprise_data/api/v1/views/lpr_data_source_snowflake.py` | Added `SnowflakeCoursePassingGradeSource`; updated `SnowflakeCourseProgressSource` with per-pair caching and pair-filtered SQL; restored `LPR_SNOWFLAKE_WAREHOUSE` / `LPR_SNOWFLAKE_ROLE` support |
| `enterprise_data/api/v1/views/enterprise_learner.py` | Added `_enrich_course_passing_grade_rows()` enrichment call in `list()` |
| `enterprise_data/api/v1/serializers.py` | Added `course_passing_grade` field to enrollment serializer |
| `enterprise_data/cache/__init__.py` | Removed unused `get_many` / `set_many` helpers; kept single-key `get` / `set` wrapping `TieredCache` |
| `enterprise_data/tests/lpr/test_lpr_data_source_snowflake.py` | Full test coverage for both source classes (34 tests) |
| `enterprise_data/tests/test_clients.py` | Removed erroneously added passing-grade methods; pylint fixes |

## Related Tickets

- **[ENT-11788](https://2u-internal.atlassian.net/browse/ENT-11788)** — Add cache to course-progress and add course passing grade enrichment *(this PR)*
- **[ENT-11722](https://2u-internal.atlassian.net/browse/ENT-11722)** — Add Course Passing Grade column to LPR UI and CSV in Admin Portal *(follow-up)*

Test screen shots using postgresql as datasource:
API Result:
image
UI Screen:
image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+
+[10.22.5] - 2026-04-27
+-----------------------
+  * feat: add passing grade column plus per-enrollment progress caching in learner progress report v1 API and CSV export
+
 [10.22.4] - 2026-04-27
 -----------------------
   * chore: bump package version to 10.22.4 to fix failed PyPI publish for 10.22.3 (version was not bumped in __init__.py before tagging)
--- a/enterprise_data/api/v1/views/enterprise_learner.py
+++ b/enterprise_data/api/v1/views/enterprise_learner.py
@@ -181,15 +181,15 @@ class EnterpriseLearnerEnrollmentViewSet(EnterpriseViewSetMixin, viewsets.ReadOn
             # Deduplicate here so we pass a clean list to the source (which also
             # deduplicates internally, but being explicit avoids unnecessary work).
             courseruns = list(dict.fromkeys(
-                row.get('courserun_key', '').strip()
+                (row.get('courserun_key') or '').strip()
                 for row in rows
-                if row.get('courserun_key', '').strip()
+                if (row.get('courserun_key') or '').strip()
             ))
             if not courseruns:
                 return rows
             grades = SnowflakeCoursePassingGradeSource().get_passing_grade_map(courseruns)
             for row in rows:
-                courserun = row.get('courserun_key', '').strip()
+                courserun = (row.get('courserun_key') or '').strip()
                 if courserun and courserun in grades:
                     row['course_passing_grade'] = grades[courserun]
             return rows
--- a/enterprise_data/api/v1/views/enterprise_learner.py
+++ b/enterprise_data/api/v1/views/enterprise_learner.py
@@ -29,7 +29,7 @@ from enterprise_data.renderers import EnrollmentsCSVRenderer
 from enterprise_data.utils import subtract_one_month
 
 from .base import EnterpriseViewSetMixin
-from .lpr_data_source_snowflake import SnowflakeCourseProgressSource
+from .lpr_data_source_snowflake import SnowflakeCoursePassingGradeSource, SnowflakeCourseProgressSource
 
 LOGGER = getLogger(__name__)
 
@@ -71,6 +71,7 @@ class EnterpriseLearnerEnrollmentViewSet(EnterpriseViewSetMixin, viewsets.ReadOn
         'is_subsidy', 'course_product_line', 'budget_id',
         'enterprise_flex_group_name', 'enterprise_flex_group_uuid',
         'course_progress',
+        'course_passing_grade',
     ]
 
     # TODO: Remove after we release the streaming csv changes
@@ -83,9 +84,9 @@ class EnterpriseLearnerEnrollmentViewSet(EnterpriseViewSetMixin, viewsets.ReadOn
         """
         Returns all learner enrollment records for a given enterprise.
 
-        ``course_progress`` is not a column on ``EnterpriseLearnerEnrollment``,
-        so we add a synthetic placeholder column here and enrich it later from
-        Snowflake in the response path.
+        ``course_progress`` and ``course_passing_grade`` are not columns on
+        ``EnterpriseLearnerEnrollment``, so we add synthetic placeholder columns
+        here and enrich both later from Snowflake in the response path.
         """
         if getattr(self, 'swagger_fake_view', False):
             # queryset just for schema generation metadata
@@ -96,11 +97,15 @@ class EnterpriseLearnerEnrollmentViewSet(EnterpriseViewSetMixin, viewsets.ReadOn
 
         # TODO: Created a ticket ENT0-9531 to add the cache on this viewset
 
-        # Add a synthetic placeholder column so the serialized response shape
-        # always includes `course_progress`; real values are merged in later.
+        # Add synthetic placeholder columns so the serialized response shape
+        # always includes `course_progress` and `course_passing_grade`; real
+        # values are merged in later from Snowflake during enrichment.
         enrollments = EnterpriseLearnerEnrollment.objects.filter(
             enterprise_customer_uuid=enterprise_customer_uuid
-        ).extra(select={'course_progress': 'NULL'})
+        ).extra(select={
+            'course_progress': 'NULL',
+            'course_passing_grade': 'NULL',
+        })
         enrollments = self.apply_filters(enrollments)
 
         return enrollments
@@ -108,7 +113,8 @@ class EnterpriseLearnerEnrollmentViewSet(EnterpriseViewSetMixin, viewsets.ReadOn
     def list(self, request, *args, **kwargs):
         """
         Override the list method to handle streaming CSV download and enrich
-        the ``course_progress`` field from Snowflake's internal LPR table.
+        the ``course_progress`` and ``course_passing_grade`` fields from
+        Snowflake's internal LPR table and course-overviews table respectively.
         """
         if request.accepted_renderer.format == 'csv':
             return StreamingHttpResponse(
@@ -118,7 +124,7 @@ class EnterpriseLearnerEnrollmentViewSet(EnterpriseViewSetMixin, viewsets.ReadOn
             )
 
         response = super().list(request, *args, **kwargs)
-        self._enrich_course_progress(response)
+        self._enrich_lpr_fields(response)
         return response
 
     def _enrich_course_progress_rows(self, rows):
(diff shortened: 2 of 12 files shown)


enterprise_data/api/v1/views/lpr_data_source_base.py
```diff
@@ -31,6 +31,7 @@ class LPRSerializerShapeMixin:
         'enterprise_customer_uuid', 'enterprise_sso_uid', 'created', 'course_api_url',
         'total_learning_time_hours', 'is_subsidy', 'course_product_line', 'budget_id',
         'enterprise_flex_group_name', 'enterprise_flex_group_uuid',
+        'course_passing_grade',
```
iloveagent57: This is different from the order in the serializer and renderer - those classes have `course_passing_grade` at the very end. I'm not super familliar with this repo, so I'm not sure if it truly matters, but it would be nice if the ordering was the same across all of these classes.

CHANGELOG.rst
```diff
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[10.22.5] - 2026-04-27
+-----------------------
+  * feat: add support for Snowflake in learner progress report v1 API and CSV export
```
iloveagent57: I think this is a miscopy, should be something like "passing grade column plus per-enrollment progress caching".

enterprise_data/api/v1/views/lpr_data_source_snowflake.py
```diff
@@ -40,119 +50,432 @@
...
+            )
+
+    # ------------------------------------------------------------------
+    # Connection
+    # ------------------------------------------------------------------
+
+    @staticmethod
+    def _get_connection():
+        """Delegate to the module-level Snowflake connection factory."""
+        return _get_snowflake_connection()
```
iloveagent57: You should consider caching this in an instance variable to avoid extra connection overhead, e.g.
```python
if not getattr(self, '_connection', None):
    self._connection = _get_snowflake_connection()
return self._connection
```

enterprise_data/cache/__init__.py
```diff
@@ -57,3 +59,58 @@ def set(key, value, timeout=DEFAULT_TIMEOUT):  # pylint: disable=redefined-built
...
+    """
+    cached_values = {}
+    missing_keys = []
+
+    for key in keys:
+        cached_response = TieredCache.get_cached_response(key)
+        if cached_response.is_found:
+            cached_values[key] = cached_response.value
+        else:
+            missing_keys.append(key)
```
iloveagent57: Because this function is expected to be fetching many cache keys in each call, I would actually not bother with TieredCache for this use-case and go straight to the django cache to make use of the more efficient underlying `django_cache.get_many()`.

enterprise_data/cache/__init__.py
```diff
@@ -57,3 +59,58 @@ def set(key, value, timeout=DEFAULT_TIMEOUT):  # pylint: disable=redefined-built
...
+
+    if not missing_keys:
+        return cached_values
+
+    django_cached_values = django_cache.get_many(missing_keys)
+    for key, value in django_cached_values.items():
+        DEFAULT_REQUEST_CACHE.set(key, value)
```
iloveagent57: ...which means you could skip this too. The other nice part about not using TieredCache is that the mirrored `set_many()` below already doesn't use it.

enterprise_data/tests/lpr/test_lpr_data_source_snowflake.py
```diff
@@ -1,127 +1,623 @@
...
 
-    @patch('enterprise_data.api.v1.views.lpr_data_source_snowflake.settings')
-    def test_custom_internal_table(self, mock_settings):
-        mock_settings.LPR_SNOWFLAKE_DATABASE = 'STAGING'
-        mock_settings.LPR_SNOWFLAKE_SCHEMA = 'REPORTING'
-        mock_settings.LPR_SNOWFLAKE_INTERNAL_TABLE = 'COURSE_PROGRESS'
-        assert SnowflakeCourseProgressSource._internal_table() == 'STAGING.REPORTING.COURSE_PROGRESS'
+def _passing_grade_cache_key(courserun_key):
+    """Build the expected passing-grade cache key."""
+    return get_key('lpr_course_passing_grade', courserun_key)
```
iloveagent57: nit, you could include the course overview table name in this cache key too, for parity with the progress cache key above.

enterprise_data/api/v1/views/lpr_data_source_snowflake.py
```diff
@@ -40,119 +50,432 @@
...
-            connect_kwargs['warehouse'] = warehouse
-        if role:
-            connect_kwargs['role'] = role
-        return snowflake.connector.connect(**connect_kwargs)
+
+# ---------------------------------------------------------------------------
+# SnowflakeCourseProgressSource
+# ---------------------------------------------------------------------------
+
+class SnowflakeCourseProgressSource:
```
iloveagent57: There's a lot of boilerplate shared between the two `*Source` tables. `_get_connection`, `_cache_timeout`,` _*_table`, settings lookups, observability logging — both classes repeat the shape. A small base class (or just module-level helpers as you've already started with `_get_snowflake_connection`) would reduce drift.

enterprise_data/api/v1/views/enterprise_learner.py
```diff
@@ -157,6 +161,40 @@ def _enrich_course_progress(self, response):
         """
         results = response.data.get('results', [])
         self._enrich_course_progress_rows(results)
+        # Also enrich passing grade per courserun
+        try:
+            self._enrich_course_passing_grade_rows(results)
+        except Exception:  # pylint: disable=broad-exception-caught
+            LOGGER.warning('Could not enrich course_passing_grade from Snowflake', exc_info=True)
```
iloveagent57: nit: this function definition is already wrapped in a broad try/except block, you can just call the function without another try/except here.

enterprise_data/api/v1/views/lpr_data_source_snowflake.py
```diff
@@ -40,119 +50,432 @@
...
+        # Fetch missing from Snowflake if needed
+        fetched_rows = 0
+        cache_payloads = {}
+        skipped_caches = {}
+        if missing_pairs:
+            fetched = self._fetch_progress_for_enrollments(enterprise_customer_uuid, missing_pairs)
+            set_entries = {}
+            for idx, pair in enumerate(missing_pairs):
+                val = fetched.get(pair)
+                progress_map[pair] = val if val is not None else None
```
iloveagent57: `val if val is not None else None` is the same as just`val`

iloveagent57 (changes requested):
This looks good overall, just a few things to clean up before merging.

enterprise_data/cache/__init__.py
```diff
@@ -51,9 +51,68 @@ def set(key, value, timeout=DEFAULT_TIMEOUT):  # pylint: disable=redefined-built
...
+    real cached value. If that behavior changes, negative-cache entries will silently
+    stop working.
+
+    Arguments:
+        mapping (dict): ``{cache_key: value}`` pairs to store.  Values can be any
+            object, including ``None`` for negative-cache entries.
+        timeout (int): Cache timeout in seconds applied to every entry.
+    """
+    for key, value in mapping.items():
+        TieredCache.set_all_tiers(key, value, django_cache_timeout=timeout)
```
iloveagent57: We should not bother with `TieredCache` at all here and just use the django cache's `set_many` directly: https://docs.djangoproject.com/en/6.0/topics/cache/#django.core.cache.cache.set_many

That will be much more efficient.

enterprise_data/cache/__init__.py
```diff
@@ -51,9 +51,68 @@ def set(key, value, timeout=DEFAULT_TIMEOUT):  # pylint: disable=redefined-built
...
+    Returns:
+        dict: Maps each key that is present in the cache to its cached value
+        (which may be ``None`` for a negative-cache hit).  Keys absent from
+        the cache are not present in the returned dict.
+    """
+    result = {}
+    for key in keys:
+        response = TieredCache.get_cached_response(key)
+        if response.is_found:
+            result[key] = response.value
```
iloveagent57: Same idea here, use `get_many` directly: https://docs.djangoproject.com/en/6.0/topics/cache/#django.core.cache.cache.get_many
> There’s also a get_many() interface that only hits the cache once. get_many() returns a dictionary with all the keys you asked for that actually exist in the cache (and haven’t expired).

So the result is already filtered to only the keys that actually exist in the cache.

iloveagent57 (approved):

PR #697, opened 2026-07-21
fix: Align learning hours calculation between engagement metrics and …

jdhanenkula:
## Description
 
This PR aligns the Learning Hours calculation between the Engagement Metrics card and the Leaderboard as part of ENT-11979, ensuring both use the same aggregation logic and return consistent values.
 
## Changes
 
- Updated `enterprise_data/admin_analytics/database/queries/fact_engagement_admin_dash.py` to align the Learning Hours calculation with the leaderboard aggregation logic.
- Updated `enterprise_data/tests/api/v1/views/test_enterprise_admin.py` to validate the updated aggregation behavior.
- Verified that the Engagement Metrics card and Leaderboard now calculate Learning Hours consistently.
- Version bump to **10.22.10** and added the corresponding **CHANGELOG** entry.
 
 
### Jira
[ENT-11979](https://2u-internal.atlassian.net/browse/ENT-11979)

## Testing
 
- Ran pytest -q  → 7 passed.
- Ran tox -e quality
image

### DIFF
--- a/enterprise_data/tests/api/v1/views/test_enterprise_admin.py
+++ b/enterprise_data/tests/api/v1/views/test_enterprise_admin.py
@@ -103,6 +103,16 @@ class TestEnterpriseAdminAnalyticsAggregatesView(JWTTestMixin, APITransactionTes
             end_date='2021-12-31',
         )
 
+    def test_learning_hours_query_uses_learner_level_rollup(self):
+        """
+        Engagement hours should match leaderboard semantics by rolling up per learner first.
+        """
+        query = self.engagement_queries.get_learning_hours_and_daily_sessions_query(self.engagement_query_filters)
+
+        assert 'GROUP BY email' in query
+        assert 'ROUND(SUM(learning_time_seconds) / 60 / 60, 1) as learning_time_hours' in query
+        assert 'ROUND(SUM(learning_time_hours), 1) as hours' in query
+
     def _mock_run_query(self, query, *args, **kwargs):
         """
         mock implementation of run_query.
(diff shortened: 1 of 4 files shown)


rgopalrao-sonata-png (approved):

enterprise_data/tests/api/v1/views/test_enterprise_admin.py
```diff
@@ -103,6 +103,16 @@ def setUp(self):
             end_date='2021-12-31',
         )
 
+    def test_learning_hours_query_uses_learner_level_rollup(self):
```
marlonkeating: This test just validates the contents of the query, not the query's logic and end results.  Can we get a more substantive test of it?

marlonkeating (changes requested):

enterprise_data/tests/api/v1/views/test_enterprise_admin.py
```diff
@@ -103,6 +104,57 @@ def setUp(self):
             end_date='2021-12-31',
         )
 
+    def test_learning_hours_rounds_each_learner_before_summing(self):
```
marlonkeating, line 107: The test code looks good as an initial approach at validating sql query behavior.  We should definitely do more of this, given that much of our unit testing around raw sql queries doesn't validate the actual behavior of the query.  

I've created a ticket [ENT-12151](https://2u-internal.atlassian.net/browse/ENT-12151) for deciding on a pattern to use with these types of unit tests going forward, so I think it would be good to add a TODO comment in this test method for refactoring to use whatever pattern we decide on once that ticket is done.

marlonkeating (approved):

PR #703, opened 2026-08-04
fix: keep non-DSC enrollments visible in the Learner Progress Report

sjasti-sonata-svg:
JIRA: https://2u-internal.atlassian.net/browse/ENT-12136

 Summary

- The `enterprise_user__is_linked=True` filter added to `EnterpriseLearnerEnrollmentViewSet.get_queryset()` to exclude unlinked learners from the LPR also silently excluded enrollments without granted Data Sharing Consent (DSC), since both cases null out the `enterprise_user` link the same way.
- Enrollments without consent should still appear in the LPR with `is_consent_granted=False`, not disappear entirely.
- Updated the filter to `Q(enterprise_user__is_linked=True) | Q(enterprise_user__isnull=True, is_consent_granted=False)`. This narrows the "show it anyway" carve-out to only the true scrubbed case (`enterprise_user` is NULL, no learner association at all) rather than any non-consented enrollment regardless of link state -- so a genuinely unlinked learner (a real `EnterpriseLearner` row with `is_linked=False`) stays excluded even if their enrollment also happens to lack consent.
- This also fixes the `overview()` analytics endpoint, since it reuses the same queryset.

 Test plan

- Updated `test_get_queryset_adds_placeholder_metadata_columns` mock assertions for the new two-step filter chain.
- Added `test_list_includes_non_consented_enrollments_with_no_enterprise_user` reproducing the reported scenario (enterprise_user is NULL).
- Added `test_list_excludes_non_consented_enrollments_of_genuinely_unlinked_learners` covering the case where a real, is_linked=False learner exists -- confirms it still stays excluded even without consent.
- Confirmed existing `test_list_excludes_enrollments_of_unlinked_learners` and `test_overview_number_of_users_excludes_unlinked_learners` still pass.

### DIFF
--- a/enterprise_data/tests/api/v1/test_views.py
+++ b/enterprise_data/tests/api/v1/test_views.py
@@ -13,6 +13,7 @@ from rest_framework import status
 from rest_framework.reverse import reverse
 from rest_framework.test import APITransactionTestCase
 
+from django.db.models import Q
 from django.utils import timezone
 
 from enterprise_data.api.v1.serializers import EnterpriseOfferSerializer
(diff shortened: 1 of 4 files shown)


enterprise_data/tests/api/v1/test_views.py
```diff
@@ -380,19 +440,22 @@ def test_get_queryset_adds_placeholder_metadata_columns(self, mock_apply_filters
...
         mock_apply_filters.return_value = filtered_enrollments
 
         result = viewset.get_queryset()
 
-        mock_filter.assert_called_once_with(
-            enterprise_customer_uuid=self.enterprise_id,
-            enterprise_user__is_linked=True,
+        mock_filter.assert_called_once_with(enterprise_customer_uuid=self.enterprise_id)
+        enrollments.filter.assert_called_once_with(
+            Q(enterprise_user__is_linked=True) | Q(enterprise_user__isnull=True, is_consent_granted=False)
```
rgopalrao-sonata-png, line 456: check one use case where enterprise_user IS NULL and is_consent_granted=False

rgopalrao-sonata-png (approved):

enterprise_data/tests/api/v1/test_views.py
sjasti-sonata-svg: Noted and it is already covered .

marlonkeating (approved):

PR #704, opened 2026-08-06
fix: simplify unlinked-learner exclusion in the Learner Progress Report

sjasti-sonata-svg:
JIRA: https://2u-internal.atlassian.net/browse/ENT-12136

Summary
Follow-up to #703. Simplifies the queryset filter in `EnterpriseLearnerEnrollmentViewSet.get_queryset()`:

```python
 before
Q(enterprise_user__is_linked=True) | Q(enterprise_user__isnull=True, is_consent_granted=False)

 after
.exclude(enterprise_user__is_linked=False)
```

 Test plan
- Updated `test_get_queryset_adds_placeholder_metadata_columns` mock assertions for the `.exclude()` call.
- Existing `test_list_includes_non_consented_enrollments_with_no_enterprise_user` and `test_list_excludes_non_consented_enrollments_of_genuinely_unlinked_learners` (from #703) both still pass unchanged, confirming behavior is identical for the ticket's real scenarios.
- Full suite (302 tests) passes.

### DIFF
--- a/enterprise_data/api/v1/views/enterprise_learner.py
+++ b/enterprise_data/api/v1/views/enterprise_learner.py
@@ -102,8 +102,8 @@ class EnterpriseLearnerEnrollmentViewSet(EnterpriseViewSetMixin, viewsets.ReadOn
         # values are merged in later from Snowflake during enrichment.
         enrollments = EnterpriseLearnerEnrollment.objects.filter(
             enterprise_customer_uuid=enterprise_customer_uuid,
-        ).filter(
-            Q(enterprise_user__is_linked=True) | Q(enterprise_user__isnull=True, is_consent_granted=False)
+        ).exclude(
+            enterprise_user__is_linked=False,
         ).extra(select={
             'course_progress': 'NULL',
             'course_passing_grade': 'NULL',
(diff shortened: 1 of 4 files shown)


rgopalrao-sonata-png (approved):

enterprise_data/api/v1/views/enterprise_learner.py
```diff
@@ -102,8 +102,8 @@ def get_queryset(self):
         # values are merged in later from Snowflake during enrichment.
         enrollments = EnterpriseLearnerEnrollment.objects.filter(
             enterprise_customer_uuid=enterprise_customer_uuid,
-        ).filter(
-            Q(enterprise_user__is_linked=True) | Q(enterprise_user__isnull=True, is_consent_granted=False)
+        ).exclude(
+            enterprise_user__is_linked=False,
```
marlonkeating, line 106: Have we confirmed that `is_consent_granted=False` users also have `is_linked=True`? The previous implementation seemed to be operating under a different assumption.

marlonkeating: The main difference functionally between the old query and the new query is that the old query guarded against showing records where `enterprise_user` is null and `is_consent_granted` is True.  I queried the production database, however, and no such records exist.  As such, this simplification shouldn't cause regressions.

marlonkeating (approved):

